PR #54, opened 2020-10-07
Override ID attribute to not be optional

paultyng:

PR #59, opened 2020-10-15
docs: add port to example `addr`

ksatirli:
This PR adds the port to the example `addr` to be consistent with the `addr` listed earlier.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -26,7 +26,7 @@ provider "boundary" {
 
 ### Required
 
-- **addr** (String, Required) The base url of the Boundary API, e.g. "http://127.0.0.1". If not set, it will be read from the "BOUNDARY_ADDR" env var.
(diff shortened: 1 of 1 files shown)


malnick:
Thanks @ksatirli - These docs are actually auto-generated. To update this, can you update the provider help text for the addr attribute in https://github.com/hashicorp/terraform-provider-boundary/blob/main/internal/provider/provider.go#L28, then run the following to generate the docs:
1. Generate following these instructions:  https://github.com/hashicorp/terraform-provider-boundary#generating-docs
1. Remove the unused `id` attribute (should have this fixed in our upstream docs generation soon!) using this Makefile target: https://github.com/hashicorp/terraform-provider-boundary/blob/main/Makefile#L44

Thanks again for the contribution!

malnick:
Per this PR, you should be able to pull `main` and not have to worry about `id` overrides: https://github.com/hashicorp/terraform-provider-boundary/pull/54

Let me know if you have any issues.

malnick (approved):

PR #63, opened 2021-01-12
ICU-972/Update to boundary 0.1.4

louisruch:
### What does this PR do

Updates terraform provider to use boundary 0.1.4
Fixes now failing unit tests

I updated the unit tests `*ResourceDestroy` function to be consistent with checking only for the resource needed and continue otherwise.  For some unit tests, we had a seemingly uneeded logline `Got unknown resource type "boundar_user"` while others did not, some strangely did a `continue` for some cases and also had a `default:continue`


louisruch:
@malnick I made the changes will still update the doc on the process.  I have not manually tested this other than running `make dev` will do some manual testing tomorrow

louisruch:
@talanknight added you as the failing tests are due to the changes made here https://github.com/hashicorp/boundary/pull/784

internal/provider/resource_account.go
```diff
@@ -163,7 +163,7 @@ func resourceAccountRead(ctx context.Context, d *schema.ResourceData, meta inter
 
 	arr, err := aClient.Read(ctx, d.Id())
 	if err != nil {
-		if apiErr := api.AsServerError(err); apiErr != nil && apiErr.Status == int32(http.StatusNotFound) {
+		if apiErr := api.AsServerError(err); apiErr != nil && apiErr.Kind == codes.NotFound.String() {
```
talanknight, line 166: For programmatic checking like this one I trust int comparisons more than string comparisons since changes in casing or typos are easier to mess things up.  Here and everywhere else in this PR.

```suggestion
		if apiErr := api.AsServerError(err); apiErr != nil && apiErr.ResponseStatus() == http.StatusNotFound {
```

talanknight (changes requested):

talanknight (approved):

PR #64, opened 2021-01-13
Turn on dependabot for actions and Go

paultyng:
Please feel free to close if this isn't how your team works, but I noticed the docs generator was out of date, this would help be notified of those types of changes.


louisruch (approved):
LGTM

PR #70, opened 2021-01-15
Dont allow password change via tf apply

louisruch:
### What does this PR do

Removes update path for account passwords during `terraform apply`

This means if a user makes a change to the default password in their `.tf` file, it will be applied to future users created, previously created users will keep their password.

```
resource "boundary_account" "users_acct" {
  for_each       = var.users
  name           = each.key
  description    = "User account for the person named ${each.key}"
  type           = "password"
  login_name     = lower(each.key)
  password       = "password"   <--------------- I.E. user changes this
  auth_method_id = boundary_auth_method.password.id
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## Next (Unreleased)
+## 1.0.0 (Unreleased)
+
+We are bumping the version of the boundary terraform provider to 1.0.0 and will release new versions of the provider at its own cadence instead of keeping it in lockstep with Boundary.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## 1.0.0 (Unreleased)
 
-We are bumping the version of the boundary terraform provider to 1.0.0 and will release new versions of the provider at its own cadence instead of keeping it in lockstep with Boundary.
+We are bumping the version of the Boundary Terraform provider to v1.0.0 and will release new versions of the provider at its own cadence instead of keeping it in lockstep with Boundary.
 
 ### Bug Fixes
 
-* During `terraform apply`, do not update existing user account passwords when the password field is updated in tf file.
+* During `terraform apply`, do not update existing user account passwords when the password field is updated in the tf file.
   ([Issue](https://github.com/hashicorp/terraform-provider-boundary/issues/71))
   ([PR](https://github.com/hashicorp/terraform-provider-boundary/pull/70))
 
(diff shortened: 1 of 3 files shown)


louisruch:
Fixes #71 
@malnick @jefferai PTAL, I discussed this with @jefferai, for now, the best solution seems to be disabling password updates via TF, if we want to allow users to store passwords in TF in the future and are happy with any security risks that imposes we can circle back.  This at least for now allows `terraform apply` to not error out

louisruch:
ping @malnick and/or @jefferai

CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
+## 1.0.0 (Unreleased)
+
+We are bumping the version of the boundary terraform provider to 1.0.0 and will release new versions of the provider at its own cadence instead of keeping it in lockstep with Boundary.
```
jefferai, line 3:
```suggestion
We are bumping the version of the Boundary Terraform provider to 1.0.0 and will release new versions of the provider at its own cadence instead of keeping it in lockstep with Boundary.
```

CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
...
+
+### Bug Fixes
+
+* During `terraform apply`, do not update existing user account passwords when the password field is updated in tf file.
```
jefferai, line 7:
```suggestion
* During `terraform apply`, do not update existing user account passwords when the password field is updated in the tf file.
```

CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
...
+  ([Issue](https://github.com/hashicorp/terraform-provider-boundary/issues/71))
+  ([PR](https://github.com/hashicorp/terraform-provider-boundary/pull/70))
+
 ## 0.1.4 (January 14, 2021)
 
-Boundary 0.1.4 release ([see boundary changelog](https://github.com/hashicorp/boundary/blob/main/CHANGELOG.md#014-20210105))
+### New and Improved
+
+Update to `Boundary API 0.0.3` and `Boundary 0.1.4` ([see boundary changelog](https://github.com/hashicorp/boundary/blob/main/CHANGELOG.md#014-20210105))
```
jefferai, line 15: I'm not sure the best way to say this but it strikes me as a bit confusing because it suggests that the Boundary API itself is versioned, which it kind of is, but not 0.0.3. Also I don't think it needs the code quotes. Maybe: "Update to the Boundary Go SDK v0.0.3". We probably don't need to say it's against Boundary 0.1.4 because the TF provider implements what the Go SDK exposes.

jefferai (approved):
Looks good pending wordsmithing.

PR #75, opened 2021-02-01
Update boundary

louisruch:
### What does this PR do

Updates provider to use Boundary SDK V0.0.2
Updates kms wrapping to v0.6.0


louisruch:
@jefferai @malnick PTAL

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.0.1 (Unreleased)
```
malnick, line 1: Is this current release that is lining up with Boundary 0.1.5 going to be 1.0.1 of the provider? If so, should we add a new unreleased section and update the date for 1.0.1?

malnick (approved):

jefferai:
Do we actually need to release?

Do we want to changelog API/SDK updates? If so, it's missing API.

PR #96, opened 2021-04-26
docs: fix examples

apsega:
Fixed some examples where required params are not included and some other typos.


malnick:
Thank you for submitting this!

Our examples are autogenerated from the schema's in the code, so I need to look into why some of these values didn't translate as expected. More soon!

apsega:
@malnick params description are ok, just the issue with examples.

louisruch:
ty

PR #97, opened 2021-04-27
see if we can get the oidc "change state" to work...

jimlambrt:


malnick (approved):
This is awesome, thanks for fixing the algo's list and adding the OIDC test controller code - incredibly useful!

PR #105, opened 2021-05-04
Add OIDC auth method and account resource

malnick:


internal/provider/resource_account.go
```diff
@@ -91,6 +88,7 @@ func setFromAccountResponseMap(d *schema.ResourceData, raw map[string]interface{
 		return err
 	}
 
+	// TODO(malnick) - remove after deprecation cycle in favor of attributes
 	switch raw["type"].(string) {
 	case accountTypePassword:
```
jimlambrt, line 93: looks like there's no way to change the password?

internal/provider/resource_account_password_test.go
```diff
@@ -0,0 +1,150 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("boundary_account_password.foo", "description", fooAccountPasswordDesc),
+					resource.TestCheckResourceAttr("boundary_account_password.foo", "name", "test"),
+					resource.TestCheckResourceAttr("boundary_account_password.foo", "type", "password"),
+					resource.TestCheckResourceAttr("boundary_account_password.foo", "login_name", "foo"),
+					resource.TestCheckResourceAttr("boundary_account_password.foo", "password", "foofoofoo"),
+					testAccCheckAccountPasswordResourceExists(provider, "boundary_account_password.foo"),
+				),
+			},
+			importStep("boundary_account_password.foo", "password"),
```
jimlambrt, line 83: why ignore the password?

internal/provider/resource_auth_method_password.go
```diff
@@ -0,0 +1,240 @@
...
+		ReadContext:   resourceAuthMethodPasswordRead,
+		UpdateContext: resourceAuthMethodPasswordUpdate,
+		DeleteContext: resourceAuthMethodPasswordDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			IDKey: {
+				Description: "The ID of the account.",
```
jimlambrt, line 34: of the auth method?

internal/provider/resource_auth_method_password_test.go
```diff
@@ -0,0 +1,127 @@
...
+resource "boundary_auth_method_password" "foo" {
+	name        = "test"
+	description = "%s"
+	type        = "password"
+	scope_id    = boundary_scope.org1.id
+	depends_on  = [boundary_role.org1_admin]
+}`, fooAuthMethodDescUpdate)
+)
+
+func TestAccAuthMethodPassword(t *testing.T) {
```
jimlambrt, line 42: can we add min login name and min password len to these tested attributes?

internal/provider/resource_auth_method_oidc.go
```diff
@@ -0,0 +1,515 @@
...
+		ReadContext:   resourceAuthMethodOidcRead,
+		UpdateContext: resourceAuthMethodOidcUpdate,
+		DeleteContext: resourceAuthMethodOidcDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			IDKey: {
+				Description: "The ID of the account.",
```
jimlambrt, line 48: ID of the auth method?

internal/provider/resource_auth_method_oidc.go
```diff
@@ -0,0 +1,515 @@
...
+		}
+
+		if m, ok := attrs[authmethodOidcMaxAgeKey]; ok {
+			maxAge := m.(json.Number)
+			maxAgeInt, _ := maxAge.Int64()
+			d.Set(authmethodOidcMaxAgeKey, maxAgeInt)
+		}
+
+		// these are set sometimes
+		sometimesString := []string{
```
jimlambrt, line 190: I believe, the api URL prefix is required, so both of these attributes should always be set.

internal/provider/resource_auth_method_oidc.go
```diff
@@ -0,0 +1,515 @@
...
+	if attrsVal, ok := raw["attributes"]; ok {
+		attrs := attrsVal.(map[string]interface{})
+
+		// these are always set
+		d.Set(authmethodOidcStateKey, attrs[authmethodOidcStateKey].(string))
+		d.Set(authmethodOidcIssuerKey, attrs[authmethodOidcIssuerKey].(string))
+		d.Set(authmethodOidcClientIdKey, attrs[authmethodOidcClientIdKey].(string))
+		d.Set(authmethodOidcClientSecretHmacKey, attrs[authmethodOidcClientSecretHmacKey].(string))
+
+		if certs, ok := attrs[authmethodOidcIdpCaCertsKey]; ok {
```
jimlambrt, line 175: nit: consider refactoring these if statements to happen inside a for loop that ranges over all the attribute keys.  For example (not a complete list of keys, but just an idea)
```
for _, k := range []string{authmethodOidcIdpCaCertsKey, authmethodOidcAllowedAudiencesKey } {
    if obj, ok := attrs[k]; ok {
        switch obj.(type) {
            case []interface{}:
                d.Set(k, obj.([]interface{}))
            case bool:
                d.Set(k, obj.(bool))
            case string:
                d.Set(k, obj.(string))
            default:
               // raise some error 
        }
    }
}
```

jimlambrt: using a for loop like this, makes it easier to check that you didn't miss any fields.

jimlambrt: I'm happy to make this change in a future PR.

internal/provider/resource_auth_method_oidc.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	if isPrimary {
+		amrr.GetResponse().Map[authmethodOidcIsPrimaryAuthMethodForScope] = true
+	}
+
+	return setFromOidcAuthMethodResponseMap(d, amrr.GetResponse().Map)
+}
+
+func resourceAuthMethodOidcUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
jimlambrt, line 392: future PR idea for both Create/Update: use a common func to build the opts.  Not complete but something like:
```
func buildOpts(d *schema.ResourceData, fields []string) []authmethods.Option {
	opts := []authmethods.Option{}

	m := map[string]func() authmethods.Option{
		NameKey: func() authmethods.Option {
			val, ok := d.GetOk(NameKey)
			if ok {
				return authmethods.WithName(val.(string))
			}
			return authmethods.DefaultName()
		},
		DescriptionKey: func() authmethods.Option {
			val, ok := d.GetOk(DescriptionKey)
			if ok {
				return authmethods.WithName(val.(string))
			}
			return nil
		},
	}
	for _, fieldName := range fields {
		if o := m[fieldName](); o != nil {
			opts = append(opts, o)
		}
	}
	return opts
}
```

jimlambrt (changes requested):
Couple minor suggestions for this PR and some ideas for future PRs.

internal/provider/resource_account.go
jefferai: This function sets from the response, which never returns a password. But more to the point, shouldn't we _always_ set attributes if they are returned, regardless of what type of account it is?

internal/provider/resource_auth_method_oidc.go
jefferai: You could have key clashes that way, you'd probably want to format it as `attributes.%s` and use k there.

jefferai: In fact arguably this is already a problem here.

internal/provider/resource_auth_method_oidc.go
```diff
@@ -0,0 +1,515 @@
...
+	d.Set(TypeKey, raw[TypeKey])
+
+	if attrsVal, ok := raw["attributes"]; ok {
+		attrs := attrsVal.(map[string]interface{})
+
+		// these are always set
+		d.Set(authmethodOidcStateKey, attrs[authmethodOidcStateKey].(string))
+		d.Set(authmethodOidcIssuerKey, attrs[authmethodOidcIssuerKey].(string))
+		d.Set(authmethodOidcClientIdKey, attrs[authmethodOidcClientIdKey].(string))
+		d.Set(authmethodOidcClientSecretHmacKey, attrs[authmethodOidcClientSecretHmacKey].(string))
```
jefferai, line 173: Will this in fact be set if the user hasn't provided it yet?

internal/provider/resource_auth_method_oidc.go
jefferai: I'm not sure if it's easy to range over values coming from TF. If it is, then naming these with an `attributes.` prefix and ranging is likely the way to go. Even if it isn't, you should probably ensure things are disambiguated by having the keys that are referencing attributes be prefixed as such.

jefferai: I guess since this is the OIDC specific provider maybe it's not a big deal. I'm not really sure.

internal/provider/resource_account_password_test.go
malnick: We don't want to save it in state.

internal/provider/resource_account_oidc.go
```diff
@@ -0,0 +1,198 @@
...
+
+const (
+	accountTypeOidc       = "password"
+	accountOidcIssuerKey  = "issuer"
+	accountOidcSubjectKey = "subject"
+)
+
+func resourceAccountOidc() *schema.Resource {
+	return &schema.Resource{
+		Description: "The account resource allows you to configure a Boundary account.",
```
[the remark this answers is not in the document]
malnick: I want to keep this, and account, as-is until we do a refactor across all resources so the descriptions have consistent language.

internal/provider/resource_auth_method_oidc.go
malnick: It'll be empty.

internal/provider/resource_account.go
jefferai: Resolving for now as dealing with arbitrary attributes is not yet on the table

jimlambrt (approved):

PR #114, opened 2021-06-01
Add credential library resource for Vault

malnick:
Adds the credential library resource for Vault: 

```
resource "boundary_credential_store_vault" "example" {
    name  =  "foo"
    description = "foo bar"
    address = 127.0.0.1"
    namespace = "default"
 }

resource "boundary_credential_library_vault" "example" {
  name  = "example"
  description = "example resource"
  credential_store_id = boundary_credential_store.example.id

  vault_path = "/my/secrets"
  vault_http_method = "POST"
  vault_request_body = <<EOF
{
  "key": "Value",
}
EOF
}
```

Built on https://github.com/hashicorp/boundary/tree/mgaffney-vault


internal/provider/resource_credential_library_vault.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if v, ok := d.GetOk(credentialLibraryVaultHttpMethodKey); ok {
+		opts = append(opts, credentiallibraries.WithVaultCredentialLibraryHttpMethod(v.(string)))
+	}
+
+	if v, ok := d.GetOk(credentialLibraryVaultHttpRequestBodyKey); ok {
+		opts = append(opts, credentiallibraries.WithVaultCredentialLibraryHttpRequestBody(v.(string)))
+	}
+
+	if v, ok := d.GetOk(credentialLibraryVaultPathKey); ok {
```
talanknight, line 130: What is the standard practice here? This field is required in the API.  In other places where there are fields that are required on create do we allow the request to go through and let the API return the error or do we try to capture it in the provider before the API request is made?

internal/provider/provider.go
```diff
@@ -54,20 +54,21 @@ func New() *schema.Provider {
...
-			"boundary_scope":                resourceScope(),
-			"boundary_target":               resourceTarget(),
-			"boundary_user":                 resourceUser(),
+			"boundary_account":                  resourceAccount(),
+			"boundary_account_password":         resourceAccountPassword(),
+			"boundary_account_oidc":             resourceAccountOidc(),
+			"boundary_auth_method":              resourceAuthMethod(),
+			"boundary_auth_method_password":     resourceAuthMethodPassword(),
+			"boundary_auth_method_oidc":         resourceAuthMethodOidc(),
+			"boundary_credential_library_vault": resourceCredentialLibraryVault(),
```
talanknight, line 63: Since libraries depend on stores I would have expected to see the work on the store to come first.  The way you have it here you'll need to create the store external to TF and then use TF for library operations.

talanknight (commented):
Looks like it is going in the right direction.

internal/provider/provider.go
```diff
@@ -54,20 +54,21 @@ func New() *schema.Provider {
...
+			"boundary_group":                    resourceGroup(),
+			"boundary_host":                     resourceHost(),
+			"boundary_host_catalog":             resourceHostCatalog(),
```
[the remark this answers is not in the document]
malnick: Ah good catch, thanks.

internal/provider/resource_credential_library_vault.go
malnick: When possible, I try to let the provider pass through the error, so we don't duplicate error login in the provider that already exists in Boundary.

internal/provider/provider.go
malnick: Yeah, I realized this part way through writing the library. I started adding the store this morning so I can finish writing tests for the library resource. They'll both ship in this PR :)

louisruch:
@malnick PTAL I exposed NewVaultTestServer in this PR https://github.com/hashicorp/boundary/pull/1321
And then made these updates to get all tests to pass
@talanknight PTAL

louisruch:
Just FYI I have now added the target cred lib Set functionality and associated tests...

internal/provider/resource_credential_store_vault.go
```diff
@@ -0,0 +1,345 @@
...
+				Description: "The Vault credential store name. Defaults to the resource name.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			DescriptionKey: {
+				Description: "The Vault credential store description.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			ScopeIdKey: {
```
talanknight, line 66: If changing the store id in the library has force new, then this field should also probably force new.

internal/provider/resource_credential_store_vault.go
```diff
@@ -0,0 +1,345 @@
...
+		opts = append(opts, credentialstores.DefaultVaultCredentialStoreClientCertificateKey())
+		v, ok := d.GetOk(credentialStoreVaultClientCertificateKeyKey)
+		if ok {
+			opts = append(opts, credentialstores.WithVaultCredentialStoreClientCertificateKey(v.(string)))
+		}
+	}
+
+	if len(opts) > 0 {
+		opts = append(opts, credentialstores.WithAutomaticVersioning(true))
+		aur, err := client.Update(ctx, d.Id(), 0, opts...)
```
talanknight, line 319: If aur has a different value for credentialStoreVaultTokenHmacKey than we have stored we should issue another update just on the token or maybe we should issue a read before the update.  Same thing for the clientCertificate Key.  Please also add a test for this.

talanknight (changes requested):
@louisruch

malnick:
Just a note that this PR is ready to go but we're waiting to rebase the go.mod on `boundary:main` when that gets merged next week to ensure our provider `main` branch isn't pointed at a Boundary feature branch.

examples/resources/boundary_target/resource.tf
```diff
@@ -47,6 +63,10 @@ resource "boundary_host_set" "foo" {
     boundary_host.foo.id,
     boundary_host.bar.id,
   ]
+
+  credential_library_ids = [
```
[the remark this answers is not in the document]
louisruch: ty

internal/provider/resource_credential_store_vault.go
louisruch: 👍

PR #118, opened 2021-06-13
Add managed group resource

malnick:


louisruch:
@malnick PTAL

malnick:
Thanks @louisruch!

PR #128, opened 2021-07-26
Fix Typo

lewiscowper:


malnick:
Thank you for catching this!

Can we use the American spelling here to reflect the spelling in the provider codebase instead: `organization`?

malnick (changes requested):
Spelling of `organisation` should be the American spelling to reflect the spelling in the codebase itself: `organization`.

Thank you for catching and submitting this!

malnick (approved):

PR #132, opened 2021-08-31
updating examples to use new grants_strings syntax

devops-rob:
Updating the examples to use grants_strings instead of grants as this no longer works.


louisruch:
@devops-rob this looks good, can you please run `make docs` locally to update the examples in the auto generated docs and then push that as well

PR #134, opened 2021-09-01
Deprecate target host_set_ids and credential_library_ids

louisruch:
### What does this PR do

- Adds support for new target fields `application_credential_source_ids` and `host_source_ids` 
- Deprecates `application_credential_library_ids` and `host_library_ids`

### DIFF
--- a/docs/resources/managed_group.md
+++ b/docs/resources/managed_group.md
@@ -18,11 +18,11 @@ The managed group resource allows you to configure a Boundary group.
 ### Required
 
 - **auth_method_id** (String) The resource ID for the auth method.
+- **filter** (String) Boolean expression to filter the workers for this managed group.
 
 ### Optional
 
 - **description** (String) The managed group description.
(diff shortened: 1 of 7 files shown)


louisruch:
@malnick @jefferai I realized that we never updated tf provider to use the new Source fields, so this adds support for the new and deprecated target fields.

docs/resources/managed_group.md
```diff
@@ -18,11 +18,11 @@ The managed group resource allows you to configure a Boundary group.
 ### Required
 
 - **auth_method_id** (String) The resource ID for the auth method.
+- **filter** (String) Boolean expression to filter the workers for this managed group.
```
louisruch, line 21: Probably shouldn't have been part of this PR but noticed `filter` incorrectly set as optional

malnick (approved):
Awesome, thanks @louisruch !

PR #159, opened 2021-11-26
Add dynamic host plugin catalog/set

jefferai:
This migrates existing host resources to static type (adding an explicit static type and deprecating the non-typed version), and creates plugin types for host catalog and host set.

Note: proper operation (specifically attribute null handling) depend on some post-0.7.1 Boundary patches. We could release without these patches in a release of Boundary but there will be some unexpected behavior around setting attributes to null/empty.


internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+			}
+		default:
+			d.Set(AttributesJsonKey, nil)
+		}
+	}
+	// Secrets stuff
+	{
+		// We do not save secrets into the state file, and they're not returned in
+		// the response
+		secretsRaw, ok := raw["secrets"]
```
talanknight, line 172: Should this be `secrets_hmac`?

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+			TypeKey: {
+				Description: "The host catalog type. Only `plugin` is supported, and is the default.",
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Default:     "plugin",
+			},
+			AttributesJsonKey: {
+				Description: `The attributes for the host catalog. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" or remove the block to clear all attributes in the host catalog.`,
+				Type:        schema.TypeString,
```
talanknight, line 80: Why not use `schema.TypeMap`?   I think we'd want the attributes defined directly in TF (unlike secrets which might be better to be passed in through a envar or path) and from a very quick look at the TypeMap it seems like it would be some of the diff checks for you.

internal/provider/resource_host_catalog_plugin_test.go
```diff
@@ -0,0 +1,367 @@
...
+					testAccCheckPluginHostCatalogResourceExists(provider, resName, expectedAttributesStatePreviouslySetNowEmpty),
+					resource.TestCheckResourceAttr(resName, DescriptionKey, testPluginHostCatalogDescriptionUpdate2),
+				),
+			},
+			importStep(resName, "secrets"),
+		},
+	})
+}
+
+func testAccCheckPluginHostCatalogResourceExists(testProvider *schema.Provider, name string, expSecrets expectedAttributesState) resource.TestCheckFunc {
```
talanknight, line 223:
```suggestion
func testAccCheckPluginHostCatalogResourceExists(testProvider *schema.Provider, name string, expAttrState expectedAttributesState) resource.TestCheckFunc {
```

internal/provider/resource_host_catalog_plugin_test.go
```diff
@@ -0,0 +1,367 @@
...
+	testPluginHostCatalogDescription        = "bar"
+	testPluginHostCatalogDescriptionUpdate  = "bar foo"
+	testPluginHostCatalogDescriptionUpdate2 = "bar foo foo"
+)
+
+type expectedAttributesState uint
+
+const (
+	expectedAttributesStatePreviouslyEmptyNowSet expectedAttributesState = iota
+	expectedAttributesStatePreviouslySetNoChange
```
talanknight, line 30: Consider documenting here that for the tests here, when we expect attributes to change, we'll also expect secrets to change.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+					}
+				},
+			},
+			SecretsJsonKey: {
+				Description: `The secrets for the host catalog. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" to clear any existing values. NOTE: Unlike "attributes_json", removing this block will NOT clear secrets from the host catalog; this allows injecting secrets for one call, then removing them for storage.`,
+				Type:        schema.TypeString,
+				Optional:    true,
+				Sensitive:   true,
+			},
+			SecretsHmacKey: {
```
talanknight, line 103: For my education, I don't see how the hmac is being used here.  Can you point me to how it is used?

internal/provider/resource_host_catalog_static.go
```diff
@@ -60,7 +59,54 @@ func resourceHostCatalog() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			ScopeIdKey: {
+				Description: "The scope ID in which the resource is created.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			TypeKey: {
```
talanknight, line 98: Why do we provide the option to change the type field in these TF resources which are explicitly for 1 specific type?

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+	if err := d.Set(NameKey, raw[NameKey]); err != nil {
+		return err
+	}
+	if err := d.Set(DescriptionKey, raw[DescriptionKey]); err != nil {
+		return err
+	}
+	if err := d.Set(ScopeIdKey, raw[ScopeIdKey]); err != nil {
+		return err
+	}
+	if err := d.Set(TypeKey, raw[TypeKey]); err != nil {
```
talanknight, line 123: Why do we store the type at all for this resource?  This can't change and for this resource type it can only be 1 value.  The user explicitly states they want this to be "plugin" by the resource type they define.

talanknight (changes requested):

vancluever (commented):
Mostly observational changes and nits. At least one thing that @talanknight mentioned. Looks like some of these are repeated patterns so any action take should likely be applied through the whole PR.

internal/provider/resource_host_catalog_plugin.go
jefferai: That is a very long conversation, more than I want to have here (so Slack or Zoom) but basically: TypeMap, despite being based on HCL and despite coming in as a `map[string]interface{}`, is not actually valid for anything other than strings. So if what you're providing isn't something that fits into `map[string]string` the behavior is more or less undefined or just flat out doesn't work.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+		CreateContext: resourceHostCatalogPluginCreate,
+		ReadContext:   resourceHostCatalogPluginRead,
+		UpdateContext: resourceHostCatalogPluginUpdate,
+		DeleteContext: resourceHostCatalogPluginDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			IDKey: {
```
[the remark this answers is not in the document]
jefferai: Good to know -- the rest of the TF provider has the ID key in all of the schemas so for now I'll put this here for consistency and can do a separate PR to sweep through and pull them out later.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+				Computed:    true, // If name is provided this will be computed
+			},
+			PluginNameKey: {
+				Description: "The name of the plugin that should back the resource. This or " + PluginIdKey + " must be defined.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Computed:    true,
+			},
+			TypeKey: {
```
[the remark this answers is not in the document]
jefferai: Just for parity with the old non-subtyped resource. I'm not sure if TF barfs if you include this and it's not part of the schema. I'll check that and if it doesn't barf I'll remove it.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+	if err := d.Set(TypeKey, raw[TypeKey]); err != nil {
+		return err
+	}
+	// Plugin stuff
+	{
+		if err := d.Set(PluginIdKey, raw[PluginIdKey]); err != nil {
+			return err
+		}
+		// Boundary doesn't currently return the plugin name in responses
```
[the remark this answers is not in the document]
jefferai: Fixed.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+				Description: "The HMAC'd secrets value returned from the server.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func setFromHostCatalogPluginResponseMap(d *schema.ResourceData, raw map[string]interface{}) error {
```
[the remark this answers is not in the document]
jefferai: The rest of the providers do it like this; I'm not sure why they were made that way but my guess is to be able to tell the difference between nil and empty. In theory we could pass both. Another thing for a cleanup pass sometime later.

internal/provider/resource_host_catalog_plugin.go
jefferai: See my reply to Chris' question. I'll test without it and remove if it's all good.

internal/provider/resource_host_catalog_plugin.go
jefferai: Fixed

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+				Computed:    true,
+			},
+			TypeKey: {
+				Description: "The host catalog type. Only `plugin` is supported, and is the default.",
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Default:     "plugin",
+			},
+			AttributesJsonKey: {
```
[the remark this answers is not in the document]
jefferai: Yes -- where the tests are passing `"null"` as the value. I added this for parity with the CLI, although I'm rethinking it now. Probably should just let the behavior of "removing the attributes block removes all attributes" be enough. I'll change that.

internal/provider/resource_host_catalog_plugin_test.go
```diff
@@ -0,0 +1,367 @@
...
+	)
+
+	initialHcl := fmt.Sprintf(projPluginHostCatalogBase, initialValuesStr)
+	update1Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate1)
+	update2Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate2)
+	update3Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate3)
+	update4Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate4)
+	update5Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate5)
+	update6Hcl := fmt.Sprintf(projPluginHostCatalogBase, valuesStrUpdate6)
+	_ = update5Hcl
```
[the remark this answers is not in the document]
jefferai: Ah, old bits for when I was disabling tests. Will remove.

internal/provider/resource_host_catalog_plugin.go
jefferai: Right now, it's only used in tests -- it lets us verify that secret creating, updating, and nulling are working properly.

In theory it can be used to raise an issue to the user if and when a read secrets hmac value does not match state, indicating it was changed outside of TF's purview. However, we wouldn't want to just blindly fail TF runs so I think we'd need some nuance there, and I didn't want to tackle that yet.

That said, it is _very_ useful for tests, as it'd be impossible to validate functionality without it.

internal/provider/resource_host_catalog_plugin_test.go
```diff
@@ -0,0 +1,367 @@
...
+		val, err := hcClient.Read(context.Background(), id)
+		if err != nil {
+			return fmt.Errorf("Got an error when reading host catalog %q: %v", id, err)
+		}
+		if val == nil {
+			return errors.New("empty val returned")
+		}
+		secretsHmac := val.GetResponse().Map[SecretsHmacKey]
+		attrs := val.GetResponse().Map["attributes"]
+		switch expSecrets {
```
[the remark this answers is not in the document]
jefferai: I think it's more a naming problem than anything else -- I was doing this only for secrets, but then changed it to have attributes and secrets tested in lock step for the same behaviors. I renamed the states but I didn't rename the variables in this function. I'll clarify it.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,413 @@
...
+				Description: `The attributes for the host catalog. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" or remove the block to clear all attributes in the host catalog.`,
+				Type:        schema.TypeString,
+				Optional:    true,
+				// If set to null in config and nothing comes from API, consider
+				// it the same. Same if config changes from empty to null.
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					switch {
+					case old == new:
```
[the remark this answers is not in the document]
talanknight: Why would go be doing the encoding?  The "new" string will be whatever the end user provides into the field wont it?  Or is this diffing newly read data from the API against what was read previously?  I thought it was the former since SchemaDiffSuppressFunc compares diffs in the Schema field (as opposed to the ResourceData).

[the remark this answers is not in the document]
jefferai: Resolving this as with the recent mini-RFC there is now an agreed upon way forward.

.circleci/config.yml
```diff
@@ -18,20 +18,8 @@ jobs:
           make install-go
           source ~/.bashrc
           go mod download
-    - run:
```
louisruch, line 21: Hmm, having some trouble getting `make test-database-up` to work locally.  I figured it was related to the changes here https://github.com/hashicorp/boundary/pull/1543/files, can dig in or chat with Tim about it in the New Year

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,627 @@
...
+			},
+			PluginIdKey: {
+				Description: "The ID of the plugin that should back the resource. This or " + PluginNameKey + " must be defined.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Computed:    true, // If name is provided this will be computed
+			},
+			PluginNameKey: {
+				Description: "The name of the plugin that should back the resource. This or " + PluginIdKey + " must be defined.",
```
louisruch, line 71: Consider setting `ConflictsWith`

louisruch (commented):
Took a quick look will still take another look once tests go green tomorrow, you should run `make docs` on this as well

.circleci/config.yml
louisruch: Ah I see you added https://github.com/hashicorp/boundary/commit/269fb94885ab361aec13b112e9c63a81a0b3a5d0#diff-702fdcc3f995a7ce572c8b1f03ccf326caf9599142c8de0317e60ba7082e2600 recently so we do not need to spin up a testing db, in that case we can just leave this out.  We should probably remove the `make test-database-up` if its no longer needed for tests

internal/provider/resource_host_catalog_plugin.go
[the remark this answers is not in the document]
talanknight: The mini rfc, if it is the one I am thinking of, discusses secrets.  This is the attributes field and is type string.  The input could be anything, even json that the user tries to hand craft themselves.  This is why I think we want to normalize the value entered here before doing the diff otherwise we may always see a diff in that situation.

talanknight (changes requested):
Looks good! My only concern is about how we compare the json attributes in the DiffSuppressFunc for plugin catalogs and sets, but that might be due to my misunderstanding of how things are stored or when the comparison happens.  Happy to discuss!

internal/provider/resource_host_catalog_plugin.go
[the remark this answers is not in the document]
jefferai: Why do you think that's not the behavior we want?

If the user changes the input, let's diff it. Do you think we should do something else?

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,627 @@
...
+						return true
+					case old == "" && new == "null":
+						return true
+					default:
+						return false
+					}
+				},
+			},
+			SecretsJsonKey: {
+				Description: `The secrets for the host catalog. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" to clear any existing values. NOTE: Unlike "attributes_json", removing this block will NOT clear secrets from the host catalog; this allows injecting secrets for one call, then removing them for storage.`,
```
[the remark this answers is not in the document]
jefferai: Will wrap it.

.circleci/config.yml
jefferai: It's needed -- that option is not used here. We do need to spin up a testing DB.

internal/provider/resource_host_catalog_plugin.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+		// Get current secrets_json value
+		secretsJson, err = parseutil.ParsePath(d.Get(SecretsJsonKey).(string))
+		if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+			return diag.Errorf("error parsing path with secrets: %v", err)
+		}
+		// Now that we have the value from the server, see if anything needs to be
+		// done
+		var diagWarning *diag.Diagnostic
+		clearExistingSecrets, sendSecretsToBoundary, diagWarning, err = calculateConfigHmacPlan(serverSecretsHmac, secretsJson, d)
```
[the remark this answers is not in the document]
jefferai: Done

internal/provider/resource_host_catalog_plugin.go
jefferai: Good idea, set!

internal/provider/resource_host_catalog_plugin.go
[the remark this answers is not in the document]
jefferai: Addressed!

.circleci/config.yml
jefferai: Oh sorry. I forgot that I set it to opt-in so this should do what's expected.

talanknight (approved):

PR #161, opened 2021-12-07
Correct invalid example

peterlindsten:
Credential library vault resource had an invalid example


louisruch (changes requested):
Hi there, thanks for the contribution.  Our docs are auto generated so the way to make this change would be to update the example in the `resource.tf` file here: https://github.com/hashicorp/terraform-provider-boundary/blob/main/examples/resources/boundary_credential_library_vault/resource.tf#L38
You can then run `make docs` or `go generate` to update the docs.

peterlindsten:
@louisruch Thanks for the detailed instruction.
Took a little bit to get the right go version to run on my setup - but I think this is correct now?

Ran the .tf file through terraform fmt, since it looked like the others were.

louisruch (approved):
Thanks again for the contribution

Issue #151, opened 2021-10-19
Keeps adding 'subject' to boundary_account_oidc resource even when the tf configuration is unchanged.


tomiles:
### Terraform Version
```
Terraform v1.0.9 on linux_amd64
provider registry.terraform.io/hashicorp/boundary v1.0.5
```
### Affected Resource(s)
- boundary_account_oidc

### Terraform Configuration Files
```hcl
## Backend ##
resource "boundary_auth_method_oidc" "oidc_azuread_xxxxx" {
  name                 = "AzureAD - xxxxx"
  description          = "OIDC auth method for AzureAD"
  scope_id             = boundary_scope.xxxxx.id
  claims_scopes        = [ "email", "profile", ]
  account_claim_maps   = [ "email=sub" ]
  is_primary_for_scope = true
  signing_algorithms   = ["RS256"]
  issuer               = "https://login.microsoftonline.com/${data.vault_generic_secret.oidc_azure_boundary.data["tenant_id"]}/v2.0"
  client_id            = data.vault_generic_secret.oidc_azure_boundary.data["client_id"]
  client_secret        = data.vault_generic_secret.oidc_azure_boundary.data["client_secret"]
  api_url_prefix       = "https://xxxxx.com"
  callback_url         = "https://xxxxx.com/v1/auth-methods/oidc:authenticate:callback"
}

## Users ##
locals {
  users = {
    "firstlastname" = {
      name  = "First Lastname"
      email = "First.Lastname@Org.com"
    }
  }
}

resource "boundary_account_oidc" "oidc_accounts" {
  for_each       = local.users
  name           = each.value["name"]
  description    = "OIDC account"
  auth_method_id = boundary_auth_method_oidc.oidc_azuread_xxxxx.id
  subject        = each.value["email"]
}
```

### Terraform Plan Output
```hcl
Terraform will perform the following actions:

  # boundary_account_oidc.oidc_accounts["firstlastname"] must be replaced
-/+ resource "boundary_account_oidc" "oidc_accounts" {
      ~ id             = "acctoidc_2JWoiO7faj" -> (known after apply)
        name           = "First Lastname"
      + subject        = "First.Lastname@Org.com" # forces replacement
        # (2 unchanged attributes hidden)
    }
```

### Expected Behavior
When the Terraform configuration has no changes in the `boundary_account_oidc` resource it should not report any actions.

### Actual Behavior
[...]

PR #177, opened 2022-02-23
Map Issuer and Subject from attributes elements

lrolaz:
Fix the Issuer and Subject mapping from the API for an OIDC Account.
I hope it's not a too quick and dirty Fix, but it works for the latest Boundary version.

This Pull Request can solve the issue https://github.com/hashicorp/terraform-provider-boundary/issues/151


louisruch:
@lrolaz thank you for the contribution!

louisruch (approved):

PR #178, opened 2022-02-26
Add validate func for auth method type key

KeisukeYamashita:
## WHY

As of now (2022-02-26)、curretly the `password` or `oidc` type is supported. 
Thus, I added a validation func to check on validation.


louisruch (approved):
Thanks you for the contribution!

PR #179, opened 2022-02-26
Add missing required `type` argument to `boundary_host` and `boundary_host_set`

KeisukeYamashita:
## WHY

This `type` argument is a required argument but missing from the examples.

### DIFF
--- a/docs/resources/host.md
+++ b/docs/resources/host.md
@@ -40,6 +40,7 @@ resource "boundary_host" "example" {
   description     = "My first host!"
   address         = "10.0.0.1"
   host_catalog_id = boundary_host_catalog.static.id
+  type            = "static"
 }
 ```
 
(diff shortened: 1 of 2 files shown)


louisruch:
@KeisukeYamashita Thank you for the contribution and helping keep Boundary documentation correct! The Boundary Terraform provider uses auto documentation generated from sources. 

Can you please update the examples [here](https://github.com/hashicorp/terraform-provider-boundary/blob/main/examples/resources/boundary_host/resource.tf#L23) (and similiarly for host-set) and then run `make docs` or `go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs` so these changes do not get reverted on the next run.

louisruch:
I ended up including this change in the latest set of [updates](https://github.com/hashicorp/terraform-provider-boundary/pull/201). Thanks for pointing it out!

PR #201, opened 2022-05-13
TF updates

louisruch:
### What does this PR do

- 3ecf3ff Fixes errors with how we deprecated `resource_account`. I was not here and perhaps we did plan on exposing attributes via tf. But the current way we are doing this is have a resource per type with type specific attributes.  This removes a deprecated warning when using the `resource_account_password` and clearly explains what users should do with `resource_account`
- 9aa514f Updates a number of dependencies including to Boundary 0.8.1
- 0f5cb98 makes updates needed to support new SDK which calls `go-kms-wrapping` V2
- 4c51244 fixes a flaky test, I am not sure what was happening with the regex, but locally it works and I have seen this failing before in CI
- a01f181 updates a number of documentation errors and run `make docs` on the latest version
- 37f3391 updates the changelog


louisruch:
@jefferai or @malnick PTAL I added a summary in the description of each commit

PR #213, opened 2022-06-01
bug(kms): Add support for all Boundary supported KMSes

louisruch:
After moving to go-kms-wrapping V2, the Boundary Terraform Provider
did not load all KMS plugins resulting in an error when trying to
create a wrapper for any type other than aead:

    Error: error reading wrappers from "recovery_kms_hcl":
    Error configuring kms: plugin is nil


jefferai (approved):
Looks fine but longer term we should have a plan to share assets between Boundary and the TF provider.

PR #216, opened 2022-06-03
bug(kms): Generate all kms plugin assets during goreleaser pre build

louisruch:


scripts/plugins.sh
```diff
@@ -3,6 +3,16 @@
 # This script builds the required plugins.
 set -e
 
+if [[ -z "$GOOS" ]]; then
```
louisruch, line 6: I need the `GOOS` and `GOARCH` to construct the file name correct for embedding. 
This means as the current script is run those need to be explicitly set when running, this means when running `make dev-with-plugins` `GOOS` and `GOARCH` need to be redundantly set i.e. `GOOS=darwin GOARCH=amd64 make dev-with-plugins`
If there is another way to get OS and ARCH let me know!

plugins/kms/const.go
```diff
@@ -1,3 +1,12 @@
...
 
-const KmsPluginPrefix = "boundary-plugin-kms-"
+import (
+	"runtime"
+	"strings"
+)
+
+const kmsPluginPrefix = "boundary-plugin-kms"
+
+func KmsPluginPrefix() string {
```
jefferai, line 10: Rather than doing this, you could also just change the build script to put each plugin in a separate directory.

This would also get rid of the need for globbing in the embed line.

jefferai: Also it might be cleaner to leave this as a const defined in the individual files as opposed to a function 🤷 It doesn't super matter and we can change it later...

.gitignore
```diff
@@ -35,4 +35,14 @@ website/vendor
...
+/plugins/kms/assets/darwin/arm64/boundary-plugin*
+/plugins/kms/assets/linux/386/boundary-plugin*
+/plugins/kms/assets/linux/amd64/boundary-plugin*
+/plugins/kms/assets/linux/arm/boundary-plugin*
+/plugins/kms/assets/linux/arm64/boundary-plugin*
+/plugins/kms/assets/windows/386/boundary-plugin*
+/plugins/kms/assets/windows/amd64/boundary-plugin*
+/plugins/kms/assets/freebsd/386/boundary-plugin*
+/plugins/kms/assets/freebsd/amd64/boundary-plugin*
+/plugins/kms/assets/freebsd/arm/boundary-plugin*
```
jefferai, line 48: Haven't tried it, but does `/plugins/kms/assets/*/*/boundary-plugin*` work? Not sure how greedy the globbing is in gitignores.

jefferai (approved):
LGTM!

PR #270, opened 2022-09-21
fix: set target credentials on update

ddebko:
### Summary:

When updating a target with different credential sources, the credential types that had changed from the state file were applied. Since we are using the SetCredentialSources function, any credential ids that are stored in the state file, but have not changed would be lost.

### Fix:

Gather all the credential ids when performing a SetCredentialSource action.


johanbrandhorst (commented):
Could we write a test that was broken before this fix, to prevent regressions?

johanbrandhorst (approved):
Good to go :)

louisruch (approved):
LGTM

PR #271, opened 2022-09-21
feat: add support for json credentials

ddebko:
### Summary:

Added support for static json credential type.


internal/provider/resource_credential_json_test.go
```diff
@@ -0,0 +1,196 @@
...
+	url := tc.ApiAddrs()[0]
+
+	res := jsonCredResource(
+		jsonCredName,
+		jsonCredDesc,
+		jsonCredObj,
+	)
+
+	resUpdate := jsonCredResource(
+		jsonCredName,
```
louisruch, line 60: Would be good to update the name and description and validate they both are updated in the update as well.

louisruch (approved):
One minor comment around the test otherwise this looks GREAT!

PR #272, opened 2022-09-21
fix: set secret variables to senitive

ddebko:
### Summary:

Setting variables that are secrets to sensitive so they are not printed out to the terminal


johanbrandhorst (approved):

louisruch (changes requested):
You need to run `make docs`

louisruch (approved):
LGTM

PR #273, opened 2022-09-21
fix: remove deprecated flag application_credential_sources_ids

ddebko:
### Summary:

Removing the deprecated flag application_credential_sources_ids from the target resource.


louisruch (approved):
LGTM

PR #277, opened 2022-09-27
bug(vault): Fix panic after token expires

louisruch:
Created a Vault store, expired it in Vault and waited for Boundary to expire to token. I then ran `terraform apply` which lead to the panic:

```
anic: interface conversion: interface {} is nil, not string
2022-09-27T16:48:09.875-0700 [DEBUG] plugin.terraform-provider-boundary:
2022-09-27T16:48:09.875-0700 [DEBUG] plugin.terraform-provider-boundary: goroutine 140 [running]:
2022-09-27T16:48:09.875-0700 [DEBUG] plugin.terraform-provider-boundary: github.com/hashicorp/terraform-provider-boundary/internal/provider.setFromVaultCredentialStoreResponseMap(0xc000209310?, 0x4fc9188?, 0x1)
2022-09-27T16:48:09.875-0700 [DEBUG] plugin.terraform-provider-boundary: 	/Users/louisruch/terraform-provider-boundary/internal/provider/resource_credential_store_vault.go:146 +0x525
```

Rebuilt with the new code changes and it did not panic but gave the following warning:

Screen Shot 2022-09-27 at 4 50 31 PM

### DIFF
--- a/internal/provider/resource_credential_store_vault.go
+++ b/internal/provider/resource_credential_store_vault.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/hashicorp/boundary/api"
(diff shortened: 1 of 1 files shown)


louisruch:
@johanbrandhorst @ddebko This fixes a panic that was occurring because the Vault Token HMAC is not returned for expired tokens, only the `token_status` is returned. I added details of my test to the description

johanbrandhorst (approved):
Nice work!

PR #280, opened 2022-10-04
Support for defining plugin directory

louisruch:


internal/provider/provider.go
```diff
@@ -116,17 +117,35 @@ func providerAuthenticate(ctx context.Context, d *schema.ResourceData, md *metaD
...
-			configutil.WithPluginOptions(
-				pluginutil.WithPluginsMap(kms_plugin_assets.BuiltinKmsPlugins()),
-				pluginutil.WithPluginsFilesystem(kms_plugin_assets.KmsPluginPrefix, kms_plugin_assets.FileSystem()),
-			))
+
+		opts := []pluginutil.Option{
+			pluginutil.WithPluginsMap(kms_plugin_assets.BuiltinKmsPlugins()),
+			pluginutil.WithPluginsFilesystem(kms_plugin_assets.KmsPluginPrefix, kms_plugin_assets.FileSystem()),
+		}
+		if v, exist := os.LookupEnv("BOUNDARY_PLUGIN_EXECUTION_DIR"); exist {
```
jefferai, line 125: It'd be nice if this was a part of the provider configuration too instead of a hard-to-find env var. But that could come later if need be.

jefferai (approved):
Approved with some follow on work

PR #287, opened 2022-10-12
Support Vault Credential Library Mapping Overrides

ddebko:
### Summary:

- support using `username_attribute` & `password_attribute` for vault credential library that is a username_password credential type
- support using `username_attribute`, `private_key_attribute` & `private_key_passphrase_attribute` for vault credential library that is a ssh_private_key credential type

### Open Questions:

when using the cli to remove an existing mapping override from a vault credential, you need to set the attribute value to null. Example: 
```
boundary credential-libraries update vault \
    -id clvlt_WC6cZR7gKG \
    -credential-mapping-override "username_attribute=null" \
    -credential-mapping-override "password_attribute=null"
```

I thought that it might be weird to set an attribute value to null in the terraform resource. I think people would naturally try to just remove the resource `credential_mapping_overrides` or an attribute within the map. So I added logic on the update function to handle setting a removed value from the state file to nil for the user.


louisruch (approved):
This is great thanks!

PR #289, opened 2022-10-14
fix: save resource state when failing on subtasks

ddebko:
### Summary:

The following resources have one or more subtasks when applying the terraform creation task: group, host set, role, target, user. For example, when creating a target the following steps are invoked:
1. Create the target
2. Associate host sources to the target
3. Associate credential sources to the target

Each step is a different call to the boundary API. Currently there is a bug where if I provide an invalid credential source to the target the state file does not save any information about the target. When resolving the error in the terraform source file and then reapplying the changes, boundary will return the following error: `unique name violation`. This is because we only set the state at the end of the creation function and the state is never set if the happy path is not followed.

### Solution:

Invoke the resource `setFrom{Resource Name}ResponseMap()` function when a subtasks fails. This allows us to correctly sync the resource's state to what we have in Boundary. I also validated the changes by building the provider locally and testing it out against a boundary dev instance.

### DIFF
--- a/internal/provider/resource_group.go
+++ b/internal/provider/resource_group.go
@@ -75,7 +75,7 @@ func setFromGroupResponseMap(d *schema.ResourceData, raw map[string]interface{})
 	return nil
 }
 
-func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) (errs diag.Diagnostics) {
 	md := meta.(*metaData)
 
 	var scopeId string
@@ -108,7 +108,12 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
 	if gcr == nil {
 		return diag.Errorf("group nil after create")
 	}
-	raw := gcr.GetResponse().Map
+	apiResponse := gcr.GetResponse().Map
+	defer func() {
+		if err := setFromGroupResponseMap(d, apiResponse); err != nil {
+			errs = append(errs, diag.FromErr(err)...)
+		}
+	}()
 
 	if val, ok := d.GetOk(groupMemberIdsKey); ok {
 		list := val.(*schema.Set).List()
@@ -118,16 +123,12 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 		gcsmr, err := grps.SetMembers(ctx, gcr.Item.Id, gcr.Item.Version, memberIds)
 		if err != nil {
-			return diag.Errorf("error setting principals on role: %v", err)
+			return diag.Errorf("error setting members on group: %v", err)
 		}
 		if gcsmr == nil {
 			return diag.Errorf("group nil after setting members")
 		}
-		raw = gcsmr.GetResponse().Map
-	}
-
-	if err := setFromGroupResponseMap(d, raw); err != nil {
-		return diag.FromErr(err)
+		apiResponse = gcsmr.GetResponse().Map
 	}
 
 	return nil
--- a/internal/provider/resource_role.go
+++ b/internal/provider/resource_role.go
@@ -94,7 +94,7 @@ func setFromRoleResponseMap(d *schema.ResourceData, raw map[string]interface{}) 
 	return nil
 }
 
-func resourceRoleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+func resourceRoleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) (errs diag.Diagnostics) {
 	md := meta.(*metaData)
 
 	var scopeId string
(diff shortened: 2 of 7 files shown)


internal/provider/resource_group.go
```diff
@@ -118,9 +118,15 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 		gcsmr, err := grps.SetMembers(ctx, gcr.Item.Id, gcr.Item.Version, memberIds)
 		if err != nil {
+			if err := setFromGroupResponseMap(d, raw); err != nil {
+				return diag.FromErr(err)
+			}
```
johanbrandhorst, line 123: What do you think about deferring this function call instead? Looks to me like we always want to call it after we've successfully called `grps.Create`.

internal/provider/resource_group.go
```diff
@@ -118,9 +118,15 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 		gcsmr, err := grps.SetMembers(ctx, gcr.Item.Id, gcr.Item.Version, memberIds)
 		if err != nil {
+			if err := setFromGroupResponseMap(d, raw); err != nil {
+				return diag.FromErr(err)
```
johanbrandhorst, line 122: This clobbers the original error, is there some way we can just log this error instead and show the original error to the user?

internal/provider/resource_group.go
ddebko: I was thinking that initially too, but @tmessi proposed that it might be better to explicitly call the set function when needed. I can go with either implementation.

internal/provider/resource_group.go
ddebko: good catch, maybe I can do something like this?
diag.Errorf("%s: failed to set resource state %s", originalErr.Error(), setStateErr.Error())

internal/provider/resource_group.go
johanbrandhorst: Not sure why having an explicit call is better, I would definitely prefer the defer.

internal/provider/resource_group.go
johanbrandhorst: Sounds good!

internal/provider/resource_group.go
```diff
@@ -76,6 +76,19 @@ func setFromGroupResponseMap(d *schema.ResourceData, raw map[string]interface{})
...
+		errs        diag.Diagnostics
+	)
+	defer func() {
+		if apiResponse == nil {
+			return
+		}
+		if err := setFromGroupResponseMap(d, apiResponse); err != nil {
+			errs = append(errs, diag.FromErr(err)...)
+		}
+	}()
```
johanbrandhorst, line 90: Putting this defer at the top is very confusing to me. Could we put it on line 124 and remove the conditional check on a capture variable? You're also going to want to add a named variable if you want to mutate the response:

```go
func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) (errs diag.Diagnostics) {
...
if gcr == nil {
	return diag.Errorf("group nil after create")
}
defer func() {
	if err := setFromGroupResponseMap(d, apiResponse); err != nil {
		errs = append(errs, diag.FromErr(err)...)
	}
}()
```

internal/provider/resource_group.go
```diff
@@ -118,19 +123,17 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 		gcsmr, err := grps.SetMembers(ctx, gcr.Item.Id, gcr.Item.Version, memberIds)
 		if err != nil {
-			return diag.Errorf("error setting principals on role: %v", err)
+			errs = append(errs, diag.Errorf("error setting members on group: %v", err)...)
+			return errs
```
johanbrandhorst, line 127: This shouldn't need to change, just returning will do the assignment, and the variable will be mutable in the deferred function
```suggestion
			return diag.Errorf("error setting members on group: %v", err)
```

internal/provider/resource_role.go
```diff
@@ -151,38 +151,38 @@ func resourceRoleCreate(ctx context.Context, d *schema.ResourceData, meta interf
...
+		}
+	}()
 
-	var diags diag.Diagnostics
 	if principalIds != nil {
 		tspr, err := rc.SetPrincipals(ctx, tcr.Item.Id, 0, principalIds, roles.WithAutomaticVersioning(true))
 		switch {
 		case err != nil:
-			diags = append(diags, diag.Diagnostic{Severity: diag.Error, Summary: "error setting principals", Detail: err.Error()})
+			return diag.Errorf("error setting principals on role: %v", err)
```
johanbrandhorst, line 165: Sorry for reviewing this bit by bit like this, but this is a change in behavior that's kind of a worse experience. I think in this function we should probably keep the old behavior of showing all errors rather than just the first one.

johanbrandhorst (commented):
😅

internal/provider/resource_role.go
ddebko: It's all good, it's just apart of the process to me. Thanks for catching these issues :)

internal/provider/resource_role.go
```diff
@@ -151,38 +151,38 @@ func resourceRoleCreate(ctx context.Context, d *schema.ResourceData, meta interf
...
+			apiResponse = tsgr.GetResponse().Map
 		}
 	}
 
-	if err := setFromRoleResponseMap(d, raw); err != nil {
-		return diag.FromErr(err)
-	}
-
-	return diags
+	return
```
johanbrandhorst, line 185: Nit: avoid the use of empty returns, they can hurt readability 😅

internal/provider/resource_group.go
```diff
@@ -118,19 +123,15 @@ func resourceGroupCreate(ctx context.Context, d *schema.ResourceData, meta inter
...
-		raw = gcsmr.GetResponse().Map
+		apiResponse = gcsmr.GetResponse().Map
 	}
 
-	if err := setFromGroupResponseMap(d, raw); err != nil {
-		return diag.FromErr(err)
-	}
-
-	return nil
+	return errs
```
johanbrandhorst, line 134: You can even return `nil` here and keep this diff even smaller. It will assign `nil` to the `errs` variable, then call the deferred function, which may or may not mutate the value before returning.

ddebko: I did a test locally and noticed that if I return nil at the end of the function, then only the value appended in the defer function is returned. With the current code I think its fine if we want to do that, but in the future if someone appends a warning level diagnostic to the slice and they do not return errs explicitly, then the warning value will not be in the slice.

johanbrandhorst: I see the concern. When I read this I was confused that we were returning `errs` here since it's untouched by the path to this code. It is unusual to return a value that we know is `nil`. I'd prefer to make this `nil`, and as you say, refactor to return `errs` if we do start assigning to it in this function.

johanbrandhorst (approved):
It's beautiful :it's-so-beautiful:

PR #293, opened 2022-11-01
Add support for boundary_self_managed_worker resource.

grantorchard:
Hey folks, initial cut of a self-managed worker resource.
A couple of things to note:
1. The test for worker-led auth is commented out as I couldn't find a way to generate a worker token on the fly.
2. I'm not getting tag values back from resourceSelfManagedWorkerRead when they are set via CLI/worker config, but haven't yet figured out why. Any pointers would be appreciated!


internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+					Type: schema.TypeMap,
+				},
+				Computed: true,
+			},
+			configTags: {
+				Description: "",
+				Type:        schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeMap,
+				},
```
johanbrandhorst, line 91: Judging by the API client, I think you want
```suggestion
				Type:        schema.TypeMap,
				Elem: &schema.Schema{
					Type: schema.TypeList,
					Elem: &schema.Schema{
						Type: schema.TypeString,
					},
				},
```

This is probably why you couldn't read these values.

internal/provider/resource_self_managed_worker_test.go
```diff
@@ -0,0 +1,175 @@
...
+// 				Check: resource.ComposeTestCheckFunc(
+// 					testAccCheckSelfManagedWorkerResourceExists(provider, "boundary_self_managed_worker.worker_led"),
+// 					resource.TestCheckResourceAttr("boundary_self_managed_worker.worker_led", "description", selfManagedWorkerDescUpdate),
+// 					resource.TestCheckResourceAttr("boundary_self_managed_worker.worker_led", "name", selfManagedWorkerNameUpdate),
+// 				),
+// 			},
+// 			importStep("boundary_self_managed_worker.worker_led"),
+// 		},
+// 	})
+// }
```
johanbrandhorst, line 89: I asked the team about this, and it seems like the controller led activation was added to make terraform easier, so it's probably okay for us to leave this as a manual test. However, maybe you could make it easy to test this manually by uncommenting this and making it conditional on an environment variable? Something like

```go
func TestSelfManagedWorkerWorkerLed(t *testing.T) {
    token := os.Getenv("BOUNDARY_TF_PROVIDER_TEST_WORKER_LED_TOKEN")
    if token == "" {
        t.Skip("Not running worker led activation test without worker led token present")
    }
}
```

johanbrandhorst (changes requested):
Thanks for this change! I think we can fix both of the issues you mentioned.

internal/provider/provider.go
```diff
@@ -92,6 +92,7 @@ func New() *schema.Provider {
 			"boundary_host_set_plugin":              resourceHostSetPlugin(),
 			"boundary_role":                         resourceRole(),
 			"boundary_scope":                        resourceScope(),
+			"boundary_self_managed_worker":          resourceSelfManagedWorker(),
```
louisruch, line 95: Hmm, I wonder if the name should be `self_managed_worker` instead of `worker`, We generally try and align the CLI/API with terraform provider and on the API/CLI we just use `worker`

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeString,
+				Computed:    false,
+				Optional:    true,
+			},
+			DescriptionKey: {
+				Description: "The description for the worker.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			createdTime: {
```
louisruch, line 63: We generally do not include the `created` time in the TF state, is there a specific reason it is needed for worker?

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Description: "When the self managed worker was created.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			version: {
```
louisruch, line 68: Similar to created above, since TF always uses the latest version I am not clear where this will be needed

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeInt,
+				Computed:    true,
+			},
+			address: {
+				Description: "The accessible address of the self managed worker.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			canonicalTags: {
+				Description: "",
```
louisruch, line 79: Can we add the description here. `Canonical` tags are the full set of tags (api + config) tags

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeMap,
+				},
+				Computed: true,
+			},
+			configTags: {
+				Description: "",
```
louisruch, line 87: Add description here stating these are the tags defined in the config file of the worker

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			TypeKey: {
+				Description: "The type of self managed worker. Will be one of PKI or KMS.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			apiTags: {
+				Description: "",
```
louisruch, line 110: Add description that these are the API defined tags on the worker

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeMap,
+				},
+				Computed: true,
```
louisruch, line 115: This should not be computed but optional as these should be set via TF if provided using the set worker tags

internal/provider/resource_self_managed_worker.go
```diff
@@ -0,0 +1,287 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			controllerGeneratedActivationToken: {
+				Description: "A single use token generated by the controller to be passed to the self-managed worker.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			TypeKey: {
+				Description: "The type of self managed worker. Will be one of PKI or KMS.",
```
louisruch, line 105: We normally separate resources based on `type` instead of having this provided as a schema, for example we have a vault and static resource for credential stores, doing this allows us to remove complexity in optional schemas. Here it would make since `KMS` workers cannot be created or even updated via the TF the best we can do is set worker tags.

louisruch (changes requested):
Thanks for jumping on this and great work so far, I left a couple of comments inline. Also please do run `make docs` or `go generate` to auto generate to documentation. It would be great if you could also add an example config under [examples](https://github.com/hashicorp/terraform-provider-boundary/tree/main/examples/resources)

internal/provider/resource_self_managed_worker.go
grantorchard: Happy to drop this, was just comparing it to what we see in the UI.

internal/provider/resource_self_managed_worker_test.go
grantorchard: That makes sense, I'll make that change.

internal/provider/provider.go
grantorchard: Happy to change this.

internal/provider/worker.go
```diff
@@ -0,0 +1,280 @@
...
+					},
+				},
+				Computed: true,
+			},
+			configTags: {
+<<<<<<< HEAD:internal/provider/worker.go
+				Description: "Tags as configured in the worker's HCL file.",
+=======
+				Description: "",
+>>>>>>> ef6f54e859b702c5a853f3f4e90737190ddf3dcc:internal/provider/resource_self_managed_worker.go
```
johanbrandhorst, line 82: This looks like a merge conflict that was left in 😅

johanbrandhorst (changes requested):
One small comment, otherwise looks good to me!

internal/provider/worker.go
grantorchard: Oops 😂

grantorchard: I’ll fix that shortly, but before we merge I’m having some issues with resolving the tag schema. It seems that the map elements should be a list, but that’s not supported in the schema. I’m not sure about the right approach to resolve that. At the moment if someone applies tags to a worker and we refresh, Terraform gets unhappy and nukes the resource from state, but doesn’t delete the resource itself. 
If you have any pointers for me I’d be very appreciative!

grantorchard: I’m happy enough to leave tags out so folks can get on with using this (if that makes sense), and then resolve that in a fast follow next week.

johanbrandhorst: I agree, we can leave it out.

grantorchard:
All changes are in, I think we are good to go.

internal/provider/worker.go
```diff
@@ -0,0 +1,278 @@
...
+			canonicalTags: {
+				// To be implemented
+				Description: "The aggregated view of worker tags and API tags.",
+				Type:        schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeList,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
```
johanbrandhorst, line 74: What do you mean by "to be implemented"? I must've misunderstood you when we chatted because I thought we'd leave this out. Where is it breaking today?

grantorchard: Sorry, I've left the schema in but skipping the d.Set for all tags. Would you rather I remove it from the schema altogether for now?

johanbrandhorst (approved):
LGTM but it'd be great to have @louisruch's thoughts as well. Could you also coordinate with @prsekar to create an issue for supporting tags later?

johanbrandhorst:
Reopening PR to see if we can trigger the CI job

johanbrandhorst:
Not sure what's up with Circle but I successfully ran the tests locally so lets merge this :)

PR #316, opened 2023-01-17
Remove scope id as a required param for worker creation

irenarindos:


docs/resources/worker.md
```diff
@@ -15,12 +15,9 @@ The resource allows you to create a self-managed worker object.
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
-### Required
-
-- `scope_id` (String) The scope for the worker.
-
 ### Optional
 
+- `scope_id` (String) The scope for the worker.
```
johanbrandhorst, line 20: Should we document what happens if this is not set? Does it get created in the global scope?

louisruch (commented):
It seems like the unit tests are not happy with this update

docs/resources/worker.md
louisruch: Curious if a `scope_id` is optional, my understanding was that it was always global? If that is the case shouldn't we just deprecate this field

irenarindos: Right now it defaults to global, though there has been discussion about allowing workers to exist at other scopes so I'm not sure if we should deprecate it?

louisruch: Na I think leave it in place if it is something we might be supporting in the future. Another option can be to set a default value of global via the provider

internal/provider/worker.go
```diff
@@ -214,7 +214,7 @@ func resourceWorkerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 			return diag.Errorf("error updating worker: %v", err)
 		}
 	}
-
+
```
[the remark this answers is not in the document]
irenarindos: Done, thanks!

PR #344, opened 2023-02-27
ci: Additional github action configuration

tmessi:
- Setting permissions on all workflows
- Adding actionlint workflow to lint actions
- Pin actions versions


johanbrandhorst (approved):
👏🏻

PR #349, opened 2023-03-01
Update target.md

stevenzamborsky:
Update docs for Target resource removing `scope_id` from `boundary_host` resources to prevent error

```
╷
│ Error: Unsupported argument
│
│   on main.tf line 79, in resource "boundary_host" "foo":
│   79:   scope_id        = boundary_scope.project.id
│
│ An argument named "scope_id" is not expected here.
```


PPacent (approved):
LGTM, this is correct

PR #364, opened 2023-03-21
fix: Allow users to set OIDC maxAge value to 0 to require immediate reauth

mikemountain:
Previously, setting 0 as a value for maxAge in an OIDC resource block would not update Boundary, as it was getting treated as a nil value in the provider. 

This fix addresses that issue, ensuring that both setting the value to 0 will require immediate reauthorisation, as well as removing the maxAge paramater defaulting the reauth time length to the TTL of the chosen OIDC provider.

### DIFF
--- a/internal/provider/resource_auth_method_oidc.go
+++ b/internal/provider/resource_auth_method_oidc.go
@@ -6,12 +6,13 @@ package provider
 import (
 	"context"
 	"encoding/json"
+	"net/http"
+
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/authmethods"
 	"github.com/hashicorp/boundary/api/scopes"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"net/http"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/provider/resource_auth_method_oidc.go
```diff
@@ -267,8 +267,12 @@ func resourceAuthMethodOidcCreate(ctx context.Context, d *schema.ResourceData, m
 	if clientSecret, ok := d.GetOk(authmethodOidcClientSecretKey); ok {
 		opts = append(opts, authmethods.WithOidcAuthMethodClientSecret(clientSecret.(string)))
 	}
-	if maxAge, ok := d.GetOk(authmethodOidcMaxAgeKey); ok {
+	// null values are not correctly recognized by the Terraform SDK, so we instead check here for maxAge value
```
johanbrandhorst, line 270: This wording is a bit confusing to me, by "null values", do we mean an integer with a value of zero? Do we mean an omitted value?

mikemountain: Null values in this context means if the value isn't provided in the state altogether - if you go from having a maxAge of 0 to removing it completely from a Terraform `.tf` config file, the state still treats maxAge as existing as 0

johanbrandhorst: OK, null or "omitted" sounds correct here then 👍🏻. I've seen "null value" used to mean "the zero value of the type", which is why I asked.

johanbrandhorst (approved):

internal/provider/resource_auth_method_oidc.go
```diff
@@ -115,9 +114,10 @@ func resourceAuthMethodOidc() *schema.Resource {
 				Optional:    true,
 			},
 			authmethodOidcMaxAgeKey: {
-				Description: "The max age to provide to the provider, indicating how much time is allowed to have passed since the last authentication before the user is challenged again.",
+				Description: "The max age to provide to the provider, indicating how much time is allowed to have passed since the last authentication before the user is challenged again. A value of 0 sets an immediate requirement to reauthenticate.",
```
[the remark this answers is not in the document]
mikemountain: Good call!

internal/provider/resource_auth_method_oidc.go
```diff
@@ -6,13 +6,12 @@ package provider
...
 	"context"
 	"encoding/json"
-	"net/http"
-
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/authmethods"
 	"github.com/hashicorp/boundary/api/scopes"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"net/http"
```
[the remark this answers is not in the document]
mikemountain: Thanks for catching this, not sure how it snuck down.

louisruch:
@mikemountain Before you merge this lets have a chat

PR #368, opened 2023-03-27
refactor: reduce some repeated code in tests

mikemountain:
There was a decent amount of repeated code between similar test suites, consolidated them into a few more common tests. 

Also fixed a flaky test in `TestAccCredentialLibraryVault`


jimlambrt (approved):
My only suggestion would be to pull all the test code into a `testing.go` so folks have one place to look/find it.

PR #385, opened 2023-04-25
feature: Use default auth method if no auth method ID is provided for provider

elimt:
Enable easier authentication by automatically setting the auth method id for the TF provider. Instead of passing the `auth_methoid_id` to the provider, the provider will set the `auth_method_id`.

This is similar to Boundary CLI using the primary authentication for a given scope: https://github.com/hashicorp/boundary/pull/2725

### Using the Primary Auth Method in Global Scope
```
provider "boundary" {
  addr                            = "http://127.0.0.1:9200"
  password_auth_method_login_name = "myuser"
  password_auth_method_password   = "passpass"
}
```

### Using the Primary Auth Method in a passed-in scope:
```
provider "boundary" {
  addr                            = "http://127.0.0.1:9200"
  password_auth_method_login_name = "myuser"
  password_auth_method_password   = "passpass"
  scope_id                        = "s_1234567890"
}
```

### Using auth_method_id (traditional)
```
provider "boundary" {
  addr                            = "http://127.0.0.1:9200"
  auth_method_id                  = "ampw_1234567890"
  password_auth_method_login_name = "myuser"
  password_auth_method_password   = "passpass"
}
```

#### Setting Auth Method ID
It sets the `auth_method_id` based on these rules:
- If an `auth_method_id` is passed in, that `auth_method_id` is used
- If there is only one auth method, it'll return it even if it's not the primary auth method
- If there are multiple auth methods, it'll only return the auth method for the primary auth method. OIDC auth method isn't supported
- If scope ID is empty or no primary auth method is found, it returns an error


internal/provider/provider.go
```diff
@@ -117,13 +127,31 @@ type metaData struct {
...
+	if scopeIdOk {
+		providerScope = scopeId.(string)
+	}
+
 	recoveryKmsHcl, recoveryKmsHclOk := d.GetOk("recovery_kms_hcl")
 	if token, ok := d.GetOk("token"); ok {
 		md.client.SetToken(token.(string))
 	}
 
+	// If auth_method_id is not set, get the default auth method ID for the given scope ID
```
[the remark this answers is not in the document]
jimlambrt: this might be easier to read if it's something like:
```
	var providerScope string
	scopeId, scopeIdOk := d.GetOk("scope_id")
	switch {
	case scopeIdOk:
		providerScope = scopeId.(string)
	default:
		providerScope = DEFAULT_PROVIDER_SCOPE
	}
```

internal/provider/provider.go
```diff
@@ -240,3 +268,40 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
 		return md, nil
 	}
 }
+
+// getDefaultAuthMethodId iterates over boundary client.List() to find the default auth method ID for the given scopeId.
+// If there is only one auth method, it'll return it even if it's not the primary auth method
+// If scope ID is empty or no primary auth method is found, it returns an error.
+func getDefaultAuthMethodId(ctx context.Context, client *authmethods.Client, scopeId, amType string) (string, error) {
```
jimlambrt, line 275: Do we want to support ldap auth methods in this func?

internal/provider/provider.go
```diff
@@ -117,13 +127,31 @@ type metaData struct {
...
+	}
+
 	recoveryKmsHcl, recoveryKmsHclOk := d.GetOk("recovery_kms_hcl")
 	if token, ok := d.GetOk("token"); ok {
 		md.client.SetToken(token.(string))
 	}
 
+	// If auth_method_id is not set, get the default auth method ID for the given scope ID
+	if !authMethodIdOk {
+		defaultAuthMethodId, err := getDefaultAuthMethodId(ctx, amClient, providerScope, PASSWORD_AUTH_METHOD_PREFIX)
```
jimlambrt, line 147: do we want to support ldap as well?

jimlambrt (changes requested):
couple questions, but overall looks great.

internal/provider/provider.go
elimt: Yeah, I think we can add support for that as well. A separate PR where we add LDAP support and also use the default primary method for that.

internal/provider/provider.go
elimt: Yeah, I think we can add support for that as well. A separate PR where we add LDAP support and also use the default primary method for that.

jimlambrt (approved):

PR #396, opened 2023-05-05
chore: Deprecate type field for boundary_account_password

elimt:
The `boundary_account_password` resource in the boundary provider has a required `type` field but the only valid value is "password." Deprecating the `type` field and setting a default value. Users will still be able to pass in a value for the `type` field. This will prevent existing configs from breaking.

 Example: The missing `type` field will default to `password`.
 
 ```
resource "boundary_account_password" "jeff" {
  auth_method_id = boundary_auth_method.password.id
  login_name     = "jeff"
  password       = "$uper$ecure"
}
 ```


ddebko (approved):
This looks great

PR #397, opened 2023-05-05
chore: update provider auth-method description

elimt:
- update provider auth-method description with information about when the default auth method is used


ddebko (approved):
This is great.

Issue #404, opened 2023-05-10
Unable to use provider 1.1.6 w/o primary password auth method


vhatsura:
We're using OIDC auth method as primary auth method and authenticate our boundary provider in terraform using `recovery_kms_hcl` configuration. After upgrading to 1.1.6 version, the setup stops working with arguing that we don't have password primary auth method.

### Terraform Version

1.3.9

### Affected Resource(s)

The entire provider

### Terraform Configuration Files

```hcl
provider "boundary" {
  addr             = var.boundary_address
  recovery_kms_hcl = <<EOT
kms "awskms" {
  purpose    = "recovery"
  key_id     = "global_recovery"
  region     = "${var.region}"
  access_key = "${var.aws_access_key}"
  secret_key = "${var.aws_secret_key}"
  kms_key_id = "${module.aws-kms.kms_boundary_recovery_key_id}"
}
EOT
}
```

### Debug Output

```text
Error: error looking up primary auth method for scope ID: 'global'. got 'amoidc_...' but the provider requires an auth method prefix of 'ampw'
│ 
│   with provider["registry.terraform.io/hashicorp/boundary"],
│   on providers.tf line 65, in provider "boundary":
│   65: provider "boundary" {
│ 
```

### Expected Behavior

The provider should works w/o errors

### Actual Behavior

`terraform apply` command failed with error provided above

omkensey:
If you remove the `key_id` line from the `awskms` config, does it start working?

elimt:
It seems the boundary provider in terraform is trying to find the default auth_method_id when you are using `recovery_kms_hcl `. Does it work when you explicitly pass in `auth_method_id` value? 

Example:

```
provider "boundary" {
  addr              = var.boundary_address
  auth_method_id = "oidc_12345567" #update to OIDC or any auth method id 
  recovery_kms_hcl = <<EOT
kms "awskms" {
  purpose    = "recovery"
  key_id     = "global_recovery"
  region     = "${var.region}"
  access_key = "${var.aws_access_key}"
  secret_key = "${var.aws_secret_key}"
  kms_key_id = "${module.aws-kms.kms_boundary_recovery_key_id}"
}
EOT
}
```

vhatsura:
> It seems the boundary provider in terraform is trying to find the default auth_method_id when you are using `recovery_kms_hcl `. Does it work when you explicitly pass in `auth_method_id` value?
> 
> Example:
> 
> ```
> provider "boundary" {
[...]

PR #405, opened 2023-05-10
fix: skip default auth method with recovery kms

elimt:
- Fix bug with getting default auth method when recovery kms is used for provider authentication.
- Add test coverage for the issue that caused the bug

### Considerations
- Set `boundary_auth_method_oidc.foo` test resource's scope to "global" to easily allow testing the default auth method
- Removed `auth_method_id`, `password_auth_method_login_name` & `password_auth_method_password` from `testConfigWithRecovery` since those fields are not needed for authenticating with [Recovery KMS](https://developer.hashicorp.com/boundary/docs/oss/installing/no-gen-resources#recovery-kms-workflow)

This should address this issue: https://github.com/hashicorp/terraform-provider-boundary/issues/404


jimlambrt (approved):

jimlambrt:
@elimt can you work with rel-eng to get a back port GitHub action added to this repo, like there is on boundary? 

https://github.com/hashicorp/boundary/blob/main/.github/workflows/backport.yml

elimt:
> @elimt can you work with rel-eng to get a back port GitHub action added to this repo, like there is on boundary?
> 
> https://github.com/hashicorp/boundary/blob/main/.github/workflows/backport.yml

TIL about the backport workflow. I'll reach out to  rel-eng

PR #406, opened 2023-05-11
fix: default auth method with recovery kms

elimt:
- Add test coverage for authenticating with recovery kms mechanism & an auth method that is not a `password` auth-method
- Skip fetching default auth method when `recovery_kms_hcl` is set or the client token isn't null

### Considerations
- Set `boundary_auth_method_oidc.foo` test resource's scope to "global" to easily allow testing the default auth method
- Removed `auth_method_id`, `password_auth_method_login_name` & `password_auth_method_password` from `testConfigWithRecovery` since those fields are not needed for authenticating with [Recovery KMS](https://developer.hashicorp.com/boundary/docs/oss/installing/no-gen-resources#recovery-kms-workflow). Using those fields also previously caused tests to pass with the bug.

This should cover test coverage for the bug raised: https://github.com/hashicorp/terraform-provider-boundary/issues/404

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -156,9 +156,10 @@ func providerAuthenticate(ctx context.Context, d *schema.ResourceData, md *metaD
 		md.client.SetToken(token.(string))
 	}
 
-	// If auth_method_id is not set, get the default auth method ID for the given scope ID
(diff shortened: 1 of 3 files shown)


ddebko (approved):
This looks great.

jimlambrt (approved):

PR #417, opened 2023-06-22
feat: add storage bucket resource

elimt:
- Create a new resource for storage bucket
- Add examples and docs for how to use new resource
- Tests added but they do not run on purpose and have not been validated yet because we cannot run tests for enterprise features yet. That will be addressed separately

## Considerations
- Even though only AWS can be supported today, still added the `pluginID` & `pluginName` fields so it's easy to extend once other plugin types are supported
- There is some logic around handling secrets sent to Boundary and validating secret HMAC changes that exist from `boundary_host_catalog_plugin` that was introduced in [PR-159](https://github.com/hashicorp/terraform-provider-boundary/pull/159). Used the same logic we've been using for that


internal/provider/resource_storage_bucket.go
```diff
@@ -0,0 +1,553 @@
...
+				Optional:      true,
+				ForceNew:      true,
+				Computed:      true, // If name is provided this will be computed
+			},
+			PluginNameKey: {
+				Description:   "The name of the plugin that should back the resource. This or " + PluginIdKey + " must be defined.",
+				Type:          schema.TypeString,
+				ConflictsWith: []string{PluginIdKey},
+				ExactlyOneOf:  []string{PluginIdKey, PluginNameKey},
+				Default:       defaultStorageBucketPluginName,
```
louisruch, line 69: I don't think there should be a default here. 
Apart from that won't this having a default actually break the behavior of providing either the Id or name? If the user provided the `id` but no `name` it would default to `aws`, would it not then return an error because both were provided?

louisruch (changes requested):
Took a quick look this morning thanks for jumping on this. I left a couple of comments but will leave it to your pod for a complete review and approval

internal/provider/resource_storage_bucket.go
elimt: That's a good point. I didn't think of that. I added a change to set the default plugin name during creation. Maybe we're better off not setting the default value. The user will have to set it 🤔

louisruch: yeah I think there should be no default type and always have the user set it.

docs/resources/storage_bucket.md
```diff
@@ -0,0 +1,75 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `bucket_name` (String) The name of the bucket within the external object store service.
+- `scope_id` (String) The scope for this storage bucket.
+- `secrets_json` (String, Sensitive) The secrets for the storage bucket. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" to clear any existing values. NOTE: Unlike "attributes_json", removing this block will NOT clear secrets from the storage bucket; this allows injecting secrets for one call, then removing them for storage.
+- `worker_filter` (String) Filters to the worker(s) that can handle requests for this storage bucket.
```
irenarindos, line 50:
```suggestion
- `worker_filter` (String) Filters to the worker(s) that can handle requests for this storage bucket. The filter must match an existing worker in order to create a storage bucket.
```

docs/resources/storage_bucket.md
```diff
@@ -0,0 +1,75 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "boundary_storage_bucket Resource - terraform-provider-boundary"
+subcategory: ""
+description: |-
+  The storage bucket resource allows you to configure a Boundary storage bucket. A storage bucket can only belong to the Global scope or an Org scope. This feature requires Boundary Enterprise or Boundary HCP.
+---
+
+# boundary_storage_bucket (Resource)
+
+The storage bucket resource allows you to configure a Boundary storage bucket. A storage bucket can only belong to the Global scope or an Org scope. This feature requires Boundary Enterprise or Boundary HCP.
```
irenarindos, line 11: I would also add a note here that currently we only support AWS S3 for storage buckets

docs/resources/storage_bucket.md
```diff
@@ -0,0 +1,75 @@
...
+### Required
+
+- `bucket_name` (String) The name of the bucket within the external object store service.
+- `scope_id` (String) The scope for this storage bucket.
+- `secrets_json` (String, Sensitive) The secrets for the storage bucket. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" to clear any existing values. NOTE: Unlike "attributes_json", removing this block will NOT clear secrets from the storage bucket; this allows injecting secrets for one call, then removing them for storage.
+- `worker_filter` (String) Filters to the worker(s) that can handle requests for this storage bucket.
+
+### Optional
+
+- `attributes_json` (String) The attributes for the storage bucket. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" or remove the block to clear all attributes in the storage bucket.
```
irenarindos, line 54: Attributes should be required since the region attribute within it is required- see https://developer.hashicorp.com/boundary/docs/configuration/session-recording/create-storage-bucket

irenarindos (changes requested):
Looks great! Just a few documentation comments.

docs/resources/storage_bucket.md
```diff
@@ -0,0 +1,75 @@
...
+  worker_filter = "\"pki\" in \"/tags/type\""
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `attributes_json` (String) The attributes for the storage bucket. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" or remove the block to clear all attributes in the storage bucket.
```
irenarindos, line 47: Can we clarify that the region attribute is the required attribute?

irenarindos (commented):
One more small thing but it's basically there!

docs/resources/storage_bucket.md
elimt: All the other required fields have `storagebucket` options. For example: [bucketName](https://github.com/hashicorp/boundary/blob/7f1d20ff3a95778bce75be8b02c4544bb1fb1995/api/storagebuckets/option.gen.go#L107).
They are all eventually set as storage bucket attributes in Boundary. Do you know why we didn't follow the same pattern with `region` field?

elimt: Maybe a question for @louisruch or @ddebko who worked on storage 👆

louisruch: Region is technically not required, it is only required for aws. While bucket name will required for all plugins.

irenarindos: @louisruch as far as documenting this in the provider- Should we keep `attributes_json` as an optional field but add a note that for AWS the region attribute is required?

louisruch: Yeah it makes sense for now to explicitly call this out. Though once we have multiple plugins it might get a bit verbose to list all required attributes per plugin

elimt: So Boundary will rather handle validation of the `region`. I can keep it optional & update the docs saying it's required when using the AWS plugin

I also looked into `TF RequiredWith` && [TF Schema ValidateFunc](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatefunc) to use the value of another field to validate if `attributes_json ` should be required but that does not seem possible because I can't read the value of other fields from `ValidateFunc`

irenarindos (approved):

PR #448, opened 2023-09-06
Add 'disable_credential_rotation' Attribute to Storage Bucket Provide…

AdamBouhmad:
Adds disable_credential_rotation attribute to the Storage Bucket docs. 

Docs link for reference: https://developer.hashicorp.com/boundary/docs/concepts/domain-model/storage-buckets#disable_credential_rotation


louisruch (changes requested):
TF provider docs are auto-generated, instead this should be updated here https://github.com/hashicorp/terraform-provider-boundary/blob/main/examples/resources/boundary_storage_bucket/resource.tf#L15 and then run `make docs` locally to update the docs

PR #460, opened 2023-09-21
cannot update a worker generated auth token

drewmullen:
closes: #469


drewmullen:
regardless of ForceNew, token needs to be stored in state to prevent perma drift

PR #474, opened 2023-10-09
add scope datasource, tests, and docs

grantorchard:
For different configuration workflows, we need to pass the scope id in as an argument. This data source allows discovery of a scope by its name which is a smoother pattern than needing to login to the UI to discover it.

Replaces #304.


louisruch (approved):
LGTM

grantorchard:
> The only thing I might recommend is that in the docs explain with a `info` block that using this data source for the global scope is unnecessary. When I was first learning boundary I got hung up on finding out the "id" of the global scope... which is "global" lol
> 
> You could probably DRY some of the config in your test w the configs in the scope resource but that memory savings is pretty minimal

Thanks for the feedback, I've added a comment on the global scope to that effect.

PR #485, opened 2023-10-12
chore: fix test workflow for external contributors

jimlambrt:
Just run tests on ubuntu-latest, so they will succeed for external contributors.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -6,7 +6,7 @@ permissions:
 
 jobs:
   actionlint:
-    runs-on: ${{ fromJSON(vars.RUNNER) }}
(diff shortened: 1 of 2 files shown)


jimlambrt:
@tmessi see [action failure for external contributor on PR #487](https://github.com/hashicorp/terraform-provider-boundary/actions/runs/6494145877)

moduli:
I was looking for other places that also use this `${{ fromJSON(vars.RUNNER) }}` runner. It looks like it's also used in `actionlint.yml`. Would we want that one updated or keep it as is?

jimlambrt:
added a fixup. Ty!

moduli (approved):

PR #486, opened 2023-10-12
workflows: add make-gen-delta to chk for changes from 'make gen'

jimlambrt:


Makefile
```diff
@@ -81,3 +79,14 @@ test-database-down:
...
+.PHONY: copywrite
+copywrite:
+	copywrite headers
+
+.PHONY: fmt
+fmt:
+	gofumpt -w $$(find . -name '*.go')
+
+.PHONY: gen
+gen: docs copywrite
```
moduli, line 92: Do we want to add `fmt` here?

.github/workflows/make-gen-delta.yml
```diff
@@ -0,0 +1,43 @@
+name: "make-gen-delta"
+on:
+  - workflow_dispatch
+  - push
+  - pull_request
```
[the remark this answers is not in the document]
jimlambrt: unfortunately, we can't.  If we remove pull then external contributors get no actions run for their PRs.

Makefile
jimlambrt: ty!

moduli: Looks like it's failing because it can't find `gofumpt`

jimlambrt: looking

johanbrandhorst (approved):

moduli (approved):

PR #489, opened 2023-10-16
Add static credentialstore datasource

AdamBouhmad:
Inspired by @grantorchard's scope datasource PR, I put together a first pass for the static cred store. Still needing to write tests & add documentation. 

The static cred store datasource currently will allow the discovery of the resource by it's name, and will additionally return back the description of the resource. 

```
data "boundary_credential_store_static" "test" {
  name     = "demo"
  scope_id = "p_jerZbKEwuL"
}

output "boundary_credential_store_id" {
  value = data.boundary_credential_store_static.test.id)
}

output "boundary_credential_store_name" {
  value = data.boundary_credential_store_static.test.name)
}

output "boundary_credential_store_description" {
  value = data.boundary_credential_store_static.test.description
}

----

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

boundary_credential_store_description = "testing!"
boundary_credential_store_id = "csst_qWsGl3OoHW"
boundary_credential_store_name = "demo"

### DIFF
--- a/internal/provider/data_source_credential_store_static.go
+++ b/internal/provider/data_source_credential_store_static.go
@@ -30,7 +30,7 @@ func dataSourceCredentialStoreStatic() *schema.Resource {
 				Required:    true,
 			},
 			DescriptionKey: {
-				Description: "The description of the retrieved scope",
+				Description: "The description of the retrieved credential store",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
--- a/internal/provider/data_source_credential_store_static.go
+++ b/internal/provider/data_source_credential_store_static.go
@@ -1,0 +1,112 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package provider
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/hashicorp/boundary/api"
+	"github.com/hashicorp/boundary/api/credentialstores"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceCredentialStoreStatic() *schema.Resource {
+	return &schema.Resource{
+		Description: "The static credential store data source allows you to discover an existing Boundary static credential store by name",
+		ReadContext: dataSourceCredentialStoreStaticRead,
+
+		Schema: map[string]*schema.Schema{
+			IDKey: {
+				Description: "The ID of the retrieved static credential store",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			NameKey: {
+				Description: "The name of the static credential store to retrieve",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/provider/data_source_credential_store_static.go [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+		scopeId = scopeIdVal.(string)
+	} else {
+		return diag.Errorf("no scope is set")
+	}
+
+	client := credentialstores.NewClient(md.client)
+
+	csl, err := client.List(ctx, scopeId, opts...)
+	if err != nil {
+		return diag.Errorf("error calling read static credential store: %v", err)
```
moduli, line 67: I believe we're technically listing here instead of reading. I do see that's the message we have for `scopes`. We may want to modify that one.
```suggestion
		return diag.Errorf("error calling list static credential store: %v", err)
```

internal/provider/data_source_credential_store_static.go
```diff
@@ -0,0 +1,112 @@
...
+		return diag.Errorf("static credential store nil after read")
+	}
+
+	if err := setFromStaticCredentialStoreResponseMap(d, srr.GetResponse().Map, false); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+func setFromStaticCredentialStoreReadResponseMap(d *schema.ResourceData, raw map[string]interface{}, fromRead bool) error {
```
moduli, line 102: nit: newline
```suggestion
}

func setFromStaticCredentialStoreReadResponseMap(d *schema.ResourceData, raw map[string]interface{}, fromRead bool) error {
```

internal/provider/data_source_credential_store_static.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+			ScopeIdKey: {
+				Description: "The scope for this credential store",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+		},
+	}
+}
+func dataSourceCredentialStoreStaticRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
moduli, line 45: nit: newline
```suggestion
}

func dataSourceCredentialStoreStaticRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```

internal/provider/provider.go
```diff
@@ -128,7 +128,8 @@ func New() *schema.Provider {
 			"boundary_worker":                                   resourceWorker(),
 		},
 		DataSourcesMap: map[string]*schema.Resource{
-			"boundary_scope": dataSourceScope(),
+			"boundary_scope":                   dataSourceScope(),
+			"boundary_credential_store_static": dataSourceCredentialStoreStatic(),
```
moduli, line 132: I'm not sure I follow how to break down these resources. What's the reason for creating a specific `boundary_credential_store_static` data source as opposed to a `boundary_credential_store` one?

PR #468, opened 2023-10-02
New Data source `data.boundary_user`

drewmullen:
closes: #464 

### Tests

I wrote tests that search for the global user and an org user (created by `boundary_user`). Important note: this new data source test will rely on the global admin being created with the same values as is currently used in test scaffolding. If you update the `login_name` or anything, these tests will fail as they're predicated on the current values.

```shell
$ make testacc TESTARGS='-run=TestAccUserDataSource_'

TF_ACC=1 go test ./... -v -run=TestAccUserDataSource_ -timeout 120m
?       github.com/hashicorp/terraform-provider-boundary        [no test files]
?       github.com/hashicorp/terraform-provider-boundary/plugins/kms    [no test files]
=== RUN   TestAccUserDataSource_basicOrgUser
--- PASS: TestAccUserDataSource_basicOrgUser (5.39s)
=== RUN   TestAccUserDataSource_globalAdminUser
--- PASS: TestAccUserDataSource_globalAdminUser (4.60s)
PASS
ok      github.com/hashicorp/terraform-provider-boundary/internal/provider      10.672s
```

## Notes

- `login_name`, `scope`, `authorized_actions`, and `parent_account_id` can be added to the resource `boundary_user` schema but I figured that should be a separate PR.
- I included some `.vscode` files that are quite useful for debugging in vscode. I have a blog write-up [here](https://dev.to/drewmullen/vscode-terraform-provider-development-setup-debugging-6bn). These were incorporated into the aws provider [developer docs](https://hashicorp.github.io/terraform-provider-aws/debugging/#use-visual-studio-code-debugging). I can rip this out if you don't want it in the repo

PR #491, opened 2023-10-19
Remove unneccsary template for user data source

drewmullen:
~~I believe its possible to provide default templates for resources and data sources that would automatically include examples (if found). This reduces the barrier for new contributors, peer reviewers and ensures a better and more consistent doc experience~~

Update - i thought i was required to set a template but those are optional. initially i create this PR to add default templates but, turns out there is a default template. When i initially wrote the `user` data source I think my examples weren't included because i wasnt naming the example file correctly - thus misunderstanding the default templates.

Given that, we no longer need the `user` template for user data source and we can mention templates as an option in the docs


moduli:
This [PR](https://github.com/hashicorp/terraform-provider-boundary/pull/468) also added a `scope.md.tmpl`. Is that file still necessary?

drewmullen:
Yes its still required bc it includes 1 extra line in the description

README.md
```diff
@@ -73,7 +73,7 @@ If you're using vscode, this provider has delve debugging baked into the [.vscod
...
+This provider uses the [terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs/) tool to generate documentation. Each resource and data source requires a set of examples [examples](https://github.com/hashicorp/terraform-provider-boundary/tree/main/examples) then you execute a binary generate the docs which are placed in [./docs](https://github.com/hashicorp/terraform-provider-boundary/tree/main/docs). Optionally you can specify your own a [templates](https://github.com/hashicorp/terraform-provider-boundary/tree/main/templates) and for a resource or data source if you want to add custom information (example: [data "boundary_scope"](https://github.com/hashicorp/terraform-provider-boundary/blob/main/templates/data-sources/scope.md.tmpl#L11))
```
moduli, line 76:
```suggestion
This provider uses the [terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs/) tool to generate documentation. Each resource and data source requires a set of examples [examples](https://github.com/hashicorp/terraform-provider-boundary/tree/main/examples) then you execute a binary to generate the docs which are placed in [./docs](https://github.com/hashicorp/terraform-provider-boundary/tree/main/docs). Optionally, you can specify your own [templates](https://github.com/hashicorp/terraform-provider-boundary/tree/main/templates) for a resource or data source if you want to add custom information (example: [data "boundary_scope"](https://github.com/hashicorp/terraform-provider-boundary/blob/main/templates/data-sources/scope.md.tmpl#L11))
```

moduli (commented):
Just some minor fixes on the README

drewmullen:
@moduli committed your exaction recommendation

@psekar fyi

moduli (approved):

PR #492, opened 2023-10-23
fix: Address conflict in `boundary_scope` data source schema

moduli:
Looking at the [boundary_scope data source schema](https://registry.terraform.io/providers/hashicorp/boundary/latest/docs/data-sources/scope), the examples suggest to use `parent_scope_id`, but the schema requires `scope_id`. 
```
data "boundary_scope" "org" {
  name     = "SecOps"
  parent_scope_id = "global"
}

data "boundary_scope" "project" {
  name     = "2111"
  parent_scope_id = data.boundary_scope.id
}


Required
- name (String) The name of the scope to retrieve.
- scope_id (String) The parent scope ID that will be queried for the scope.
```

This PR updates the schema to require `parent_scope_id` instead. Additionally, the PR includes the following changes
- renames some keys in `const.go` for consistency
- fixes an error message in `data_source_scope.go`


docs/data-sources/scope.md
```diff
@@ -31,7 +31,7 @@ data "boundary_scope" "project" {
 ### Required
 
 - `name` (String) The name of the scope to retrieve.
-- `scope_id` (String) The parent scope ID that will be queried for the scope.
+- `parent_scope_id` (String) The parent scope ID that will be queried for the scope.
```
moduli, line 34: I don't know what the implications are when changing something like this (since... I think it would break current implementations?). What's usually the guidance for something like this?

louisruch:
The issue here is actually the docs, `scope_id` is the correct field for a scope. The only time we use `parent_` prefix is in the `ScopeInfo` struct. The fix here should rather just remove the prefix in the examples here https://github.com/hashicorp/terraform-provider-boundary/blob/main/examples/data-sources/boundary_scope/data-source.tf#L4

louisruch:
Also +1 on the change to add the `Key` Suffix in global where missing

moduli:
> The issue here is actually the docs, `scope_id` is the correct field for a scope. The only time we use `parent_` prefix is in the `ScopeInfo` struct. The fix here should rather just remove the prefix in the examples here https://github.com/hashicorp/terraform-provider-boundary/blob/main/examples/data-sources/boundary_scope/data-source.tf#L4

@louisruch Addressed

elimt:
@moduli I am curious. When you tested it locally, did the datasource work with both `parent_scope_id` & `scope_id`?

moduli:
> @moduli I am curious. When you tested it locally, did the datasource work with both `parent_scope_id` & `scope_id`?

@elimt 
Running something like
```
resource "boundary_scope" "org1" {
  scope_id = "global"
  name     = "org1"
}

data "boundary_scope" "global_1" {
  depends_on = [boundary_scope.org1]
  scope_id   = "global"
  name       = "org1"
}

data "boundary_scope" "global_2" {
  depends_on      = [boundary_scope.org1]
  parent_scope_id = "global"
  name            = "global"
}
```

Resulted in this error
```
❯ terraform plan
╷
│ Error: Missing required argument
│ 
│   on test.tf line 29, in data "boundary_scope" "global_2":
│   29: data "boundary_scope" "global_2" {
│ 
│ The argument "scope_id" is required, but no definition was found.
╵
╷
│ Error: Unsupported argument
│ 
│   on test.tf line 31, in data "boundary_scope" "global_2":
│   31:   parent_scope_id = "global"
│ 
│ An argument named "parent_scope_id" is not expected here.
```

louisruch (approved):
LGTM

elimt (approved):
Thanks for addressing this 🙏

louisruch:
> @moduli I am curious. When you tested it locally, did the datasource work with both `parent_scope_id` & `scope_id`?

Well interestingly in the `setFromScopeReadResponseMap` function we only set the `id`, `name` and `description` since those are the only fields we care about for the datasource. The `ScopeId` is used as input for the list function, but it is not being mapped to any specific Boundary field within a Scope, see: https://github.com/hashicorp/terraform-provider-boundary/blob/main/internal/provider/data_source_scope.go#L66.

So technically it never needed to match the boundary implementation. The error Mike was getting in his manual test is likely because he undid his changes. But if you took the original PR as he had it it would have worked with `parent_scope_id`, because that parameter is being set by Terraform not by the response from Boundary hence no mismatch.

elimt:
> > @moduli I am curious. When you tested it locally, did the datasource work with both `parent_scope_id` & `scope_id`?
> 
> Well interestingly in the `setFromScopeReadResponseMap` function we only set the `id`, `name` and `description` since those are the only fields we care about for the datasource. The `ScopeId` is used as input for the list function, but it is not being mapped to any specific Boundary field within a Scope, see: https://github.com/hashicorp/terraform-provider-boundary/blob/main/internal/provider/data_source_scope.go#L66.
> 
> So technically it never needed to match the boundary implementation. The error Mike was getting in his manual test is likely because he undid his changes. But if you took the original PR as he had it it would have worked with `parent_scope_id`, because that parameter is being set by Terraform not by the response from Boundary hence no mismatch.

Ahh got it. Thanks for the explanation!

PR #493, opened 2023-10-24
feat: Add `boundary_account` data source

moduli:
This PR adds a `boundary_account` data source
```
data "boundary_account" "admin" {
  name           = "admin"
  auth_method_id = "ampw_1234567890"
}
```

<details>
<summary>Test</summary>

```
resource "boundary_auth_method_password" "ampw" {
  name     = "test"
  type     = "password"
  scope_id = "global"
}

resource "boundary_account_password" "acc_pw" {
  name           = "test"
  login_name     = "foo"
  password       = "foofoofoo"
  auth_method_id = boundary_auth_method_password.ampw.id
}

data "boundary_account" "acc_pw" {
  depends_on     = [boundary_account_password.acc_pw]
  name           = "test"
  auth_method_id = boundary_auth_method_password.ampw.id
}

resource "boundary_account_ldap" "acc_ldap" {
  auth_method_id = "amldap_1234567890"
  login_name     = "einstein"
  name           = "einstein"
  description    = "user account for einstein"
}

data "boundary_account" "acc_ldap" {
  depends_on     = [boundary_account_ldap.acc_ldap]
  name           = "einstein"
  auth_method_id = "amldap_1234567890"
}

output "resource_pw_id" {
  value = boundary_account_password.acc_pw.id
}

output "acc_pw_id" {
  value = data.boundary_account.acc_pw.id
}

output "acc_pw_name" {
  value = data.boundary_account.acc_pw.name
}

output "resource_ldap_id" {
  value = boundary_account_ldap.acc_ldap.id
}

output "acc_ldap_id" {
  value = data.boundary_account.acc_ldap.id
}

Outputs:

acc_ldap_id = "acctldap_nLnSfX5rgC"
acc_pw_id = "acctpw_aZ5OsuPE0k"
acc_pw_name = "test"
resource_ldap_id = "acctldap_nLnSfX5rgC"
resource_pw_id = "acctpw_aZ5OsuPE0k"
```
</details>


internal/provider/data_source_account.go
```diff
@@ -0,0 +1,118 @@
...
+	var authMethodId string
+	if authMethodIdVal, ok := d.GetOk(AuthMethodIdKey); ok {
+		authMethodId = authMethodIdVal.(string)
+	} else {
+		return diag.Errorf("no auth method ID provided")
+	}
+
+	acl := accounts.NewClient(md.client)
+
+	als, err := acl.List(ctx, authMethodId, opts...)
```
elimt, line 69: Could we pass in filter option (`withFilter`) here to reduce the amount of data we get back from Boundary. 
Eg. https://github.com/hashicorp/terraform-provider-boundary/blob/moduli-datasource-account/internal/provider/data_source_user.go#L111

internal/provider/data_source_account.go
```diff
@@ -0,0 +1,118 @@
...
+		return diag.Errorf("error calling list account: %v", err)
+	}
+	if als == nil {
+		return diag.Errorf("no accounts found")
+	}
+
+	var accountIdRead string
+	for _, accountItem := range als.GetItems() {
+		if accountItem.Name == name {
+			accountIdRead = accountItem.Id
```
elimt, line 80: So in this case if there are multiple accounts associated to the auth method, we only return the first one? 
I am asking because in [boundary.data_source_user](https://github.com/hashicorp/terraform-provider-boundary/blob/786dfc11cc76be4d3d17134b17b78addb6fcd357/internal/provider/data_source_user.go#L124), if there are multiple users, we throw an error.
We should possibly standardized how we handle multiple results for datasources unless there's a good reason to differ

internal/provider/data_source_account.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	if err := setFromAccountReadResponseMap(d, arr.GetResponse().Map); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func setFromAccountReadResponseMap(d *schema.ResourceData, raw map[string]interface{}) error {
```
elimt, line 108: Are there other useful fields that we return to anyone using this datasource?

internal/provider/data_source_account.go
moduli: I *think* there are checks where resources can't have the same name? 

I do agree we should try to standardize how we implement these. I have noticed some differences between `boundary_user` and `boundary_scope`.

elimt: yeah that's true actually. There can only be 1 unique account name. ignore my comment then

moduli: @elimt Made some changes here. Let me know what you think. Would like to apply the same paradigm to the other data sources.

internal/provider/data_source_account.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(accounts) == 0 {
+		return diag.Errorf("no matching account found: %v", err)
+	}
+
+	if len(accounts) > 1 {
+		return diag.Errorf("error found more than 1 account: %v", err)
+	}
+
+	var accountId string
+	for _, account := range accounts {
```
elimt, line 108: If we have the line above that always checks to make sure we only have 1 account (`if len(accounts) > 1`), do we still need the for loop? Won't there always be only 1 account in the list at this point?

moduli: Good point. What do you suggest here? 

We could just get rid of that block and go directly to
```
arr, err := acl.Read(ctx, accounts[0].Id)
```

elimt: For Boundary List calls where we can have multiple items, I wonder if we should also always return 1 time for datasources or a list which could have either 1 item in it or more than 1 item? Right now we follow a pattern of always returning 1 item for a data source.

Would customers come back to ask for a new datasource that can return all items of a resource based on attributes/filters passed in because they do not want to repeat calls to a datasource to get more data.

I'll leave it to Louis to share his thoughts on this.

cc: @louisruch

moduli: Good question. I had similar thoughts. Right now, the current data sources just do a search by `name`. Is that all we want here? Do we want to filter by other attributes? 

Alternatively, we could just start with this current approach for now and iterate in the future.

louisruch: As far as I am aware a datasource must always match to a single resource Boundary. This is needed so you can bring in a resource by some filter and then use the id or other values in other resources. And this will be done via tf mapping.

As for future filters I think we definitely will want to use more than just name, but its a great place to start and we can expand as needed.

internal/provider/data_source_account.go
```diff
@@ -0,0 +1,139 @@
...
+		return diag.Errorf("error calling list account: %v", err)
+	}
+	if accounts == nil {
+		return diag.Errorf("no accounts found")
+	}
+	if len(accounts) == 0 {
+		return diag.Errorf("no matching account found: %v", err)
+	}
+	if len(accounts) > 1 {
+		return diag.Errorf("error found more than 1 account: %v", err)
```
elimt, line 103: Will `len(accounts) == 0` & `len(accounts) > 1` produce errors from Boundary? I didn't think so. Asking because you always return an error in those cases

elimt (approved):
Minor comment. Otherwise looks good. Thanks 🙏

internal/provider/data_source_account.go
moduli: Chatted with Elim. I was unnecessarily including `err` in these error messages.

PR #512, opened 2023-10-30
enable support for ssh valid certs additional valid principals field

kheina:


elimt:
LGTM. Can you please add an example so it's easier for folks to use/set up

ddebko (changes requested):
There is some weird output I have never seen before from the unit tests:
```
Error: ../../../go/pkg/mod/github.com/hashicorp/boundary@v0.14.2/internal/kms/kms.go:603:37: type func(i "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion, j "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion) bool of func(i, j wrappingKms.KeyVersion) bool {…} does not match inferred type func(a "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion, b "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion) int for func(a E, b E) int
```

This might just be a conflict with versioning in the go mod?

ddebko:
This PR needs to be kept open until the next release of Boundary fixes the dependency issues with the following [package](https://pkg.go.dev/golang.org/x/exp/slices). Afterwards, this PR will need to be updated with the latest Boundary version.

psekar:
> There is some weird output I have never seen before from the unit tests:
> 
> ```
> Error: ../../../go/pkg/mod/github.com/hashicorp/boundary@v0.14.2/internal/kms/kms.go:603:37: type func(i "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion, j "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion) bool of func(i, j wrappingKms.KeyVersion) bool {…} does not match inferred type func(a "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion, b "github.com/hashicorp/go-kms-wrapping/extras/kms/v2".KeyVersion) int for func(a E, b E) int
> ```
> 
> This might just be a conflict with versioning in the go mod?

This should be fixed now

ddebko (approved):

PR #515, opened 2023-11-07
refact: Update existing data sources to use a consistent pattern

moduli:


jimlambrt (commented):
Any chance we can just get https://github.com/hashicorp/terraform-provider-boundary/pull/127 across the line vs hand writing these?

moduli:
> Any chance we can just get #127 across the line vs hand writing these?

@jimlambrt I can try to take a look at that PR. I was thinking of doing 1-2 more of these by hand to fulfill some customer requests (since it will likely be quicker to complete?), but I do think that linked PR negates the need for this one. Thoughts?

jimlambrt:
We can continue down this path of custom data sources, but I'm a bit worried we might have breaking changes when we generate them.

moduli:
> We can continue down this path of custom data sources, but I'm a bit worried we might have breaking changes when we generate them.

Fair point. I can try to dig into that PR and see what it would take.

moduli:
Closing this PR in favor of https://github.com/hashicorp/terraform-provider-boundary/pull/127.

PR #558, opened 2024-02-06
feat(policies): support storage policies within terraform

kheina:
NOTE: requires https://github.com/hashicorp/boundary/pull/4385 to be merged and tagged and go.mod updated before merging and deploying

testing was done manually due to policies being an enterprise feature


internal/provider/resource_policy.go
```diff
@@ -0,0 +1,428 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+			},
+			policyStorageDeleteAfterDaysKey: {
+				Description: "The number of days after which a session recording will be automatically deleted. Defaults to 0: never automatically delete. However, " + policyStorageDeleteAfterDaysKey + " and " + policyStorageRetainForDaysKey + " cannot both be 0.",
+				Type:        schema.TypeInt,
+				Optional:    true,
+			},
+			policyStorageDeleteAfterOverridableKey: {
+				Description: "Whether or not the associated " + policyStorageDeleteAfterDaysKey + " value can be overriden by other organizations that this policy may apply to.",
```
louisruch, line 81:
```suggestion
				Description: "Whether or not the associated " + policyStorageDeleteAfterDaysKey + " value can be overridden by org scopes.",
```

internal/provider/resource_policy.go
```diff
@@ -0,0 +1,428 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+			},
+		},
+	}
+}
+
+func parseAttributesInt(in any) (int64, bool) {
+	fmt.Printf("in: %#v", in)
```
louisruch, line 90: Remove testing log

internal/provider/resource_policy.go
```diff
@@ -0,0 +1,428 @@
...
+)
+
+const (
+	policyStorageRetainForDaysKey          = "retain_for_days"
+	policyStorageRetainForOverridableKey   = "retain_for_overridable"
+	policyStorageDeleteAfterDaysKey        = "delete_after_days"
+	policyStorageDeleteAfterOverridableKey = "delete_after_overridable"
+)
+
+func resourcePolicy() *schema.Resource {
```
louisruch, line 25: With the TF provider we have been creating sub type specific resources, some of the older types (e.g. targets) we don't do this, but that should be refactored at some point.

So this should be come `resourcePolicyStorage`

internal/provider/resource_policy.go
```diff
@@ -0,0 +1,428 @@
...
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			IDKey: {
+				Description: "The ID of the policy.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			TypeKey: {
```
louisruch, line 45: Won't be needed anymore

internal/provider/resource_policy.go
```diff
@@ -0,0 +1,428 @@
...
+		if err := d.Set(policyStorageRetainForOverridableKey, nil); err != nil {
+			return err
+		}
+		if err := d.Set(policyStorageDeleteAfterDaysKey, nil); err != nil {
+			return err
+		}
+		if err := d.Set(policyStorageDeleteAfterOverridableKey, nil); err != nil {
+			return err
+		}
+	}
```
louisruch, line 221: This should not be possible

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const (
+	policyPolicyIdKey         = "policy_id"
+	policyStoragePolicyPrefix = "pst_"
+)
+
+func resourcePolicyAttachment() *schema.Resource {
```
louisruch, line 23: Should become `resourcePolicyStorageAttachment`

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+	"strings"
+
+	"github.com/hashicorp/boundary/api"
+	"github.com/hashicorp/boundary/api/scopes"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const (
+	policyPolicyIdKey         = "policy_id"
```
louisruch, line 19: ?, why double policy

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+	policyStoragePolicyPrefix = "pst_"
```
louisruch, line 20: why prefix `policy`

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+	var policy_type string
+	switch {
+	case strings.HasPrefix(policyId, policyStoragePolicyPrefix):
+		if _, err := scp.AttachStoragePolicy(ctx, scopeId, 0, policyId, opts...); err != nil {
+			return diag.FromErr(err)
+		}
+		policy_type = "storage"
+	default:
+		return diag.Errorf("unknown policy type provided.")
+	}
```
louisruch, line 80: Not sure we need the  check here, but this should only ever support storage policy types so if you really want a check you can just do the `if strings.HasPrefix`, though if you are doing this check you should check that the scopeId is `global` or that of an org scope only as well

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if err := d.Set(policyPolicyIdKey, policyId); err != nil {
+		// TODO: does this ordering prevent tf from storing the tfstate if the attach fails?
```
louisruch, line 83: Not sure I am following this TODO

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+		return diag.FromErr(err)
+	}
+
+	// scopes can have one and only one policy of each type
```
louisruch, line 87: This should be able to be updated to scopeId with the subtype specific resource change

internal/provider/resource_policy_attachment.go
```diff
@@ -0,0 +1,167 @@
...
+	switch {
+	case strings.HasPrefix(policyId, policyStoragePolicyPrefix):
+		if policyId != s.GetItem().StoragePolicyId {
+			// this policy is no longer attached to the scope, destroy this resource
+			d.SetId("")
+			return nil
+		}
+	default:
+		return diag.Errorf("unknown policy type provided.")
+	}
```
louisruch, line 133: Not sure I am following why we are doing this, why not just set storagePolicyId as returned?

louisruch (changes requested):
You will need to run `make docs` locally, otherwise left a couple of inline comments, some of them should remove a number of subtype checks performed here.

internal/provider/resource_policy.go
kheina: ok, that makes the policy a bit more straight forward, but do we want the attachment to still be policy type agnostic? or allow for other policy types in the future?

internal/provider/resource_policy_attachment.go
kheina: I think that the policy attachment should stay subtype-agnostic

kheina: I don't think that there's a reason to need to add new policy type attachments for each new type. unlike the actual policy definition, there's no risk of schema bloat or anything with the attachment specifically.

louisruch: I do not believe we can make the assertion that future policies will be limited to one and only one. The plan is to allow multiple policies possibly of the same type, for storage we know we only want to limit it to one and only one (at least for now)

internal/provider/resource_policy_attachment.go
louisruch: On a read we should set it as this will allow TF to determine there has been a change and then reset it to what the TF file wants it to be

kheina: this is the current behavior, just not explicitly. since the resource is destroyed here, tf triggers a new create which replaces the active policy

internal/provider/resource_policy_attachment.go
louisruch: So there are three things that make me question if this can be subtype-agnostic:
- Having `boundary_policy_storage_attachment` in the Terraform file means it becomes human readable on what the attachment is, otherwise someone needs to know that the prefix indicates it is a storage type. 
- Future policies could allow multiple of that policy type to be associated with a scope
- We also cannot guarantee that other policies wont require additional fields when associating them with a scope, for example with Storage policies we have override behavior as part of the policy - but that might not always be the case.

I am not sure I see much of a benefit in locking into a subtype agnostic approach.

louisruch (changes requested):
One new comment and responded to the one thread re subtype agnostic attachments. Regardless on the final approach there we should add `scope` to the resource so `resource_scope_policy_attachment` or `resource_scope_policy_storage_attachment`

internal/provider/resource_policy_attachment.go
louisruch: Might be good to get @AdamBouhmad thoughts on this point. Which of the following do you prefer, noting some of my concerns above:

```
resource "boundary_scope_policy_attachment" "foo" {
  scope_id  = boundary_scope.org.id
  policy_id = boundary_storage_policy.boo.id
}
```
or
```
resource "boundary_scope_policy_storage_attachment" "foo" {
  scope_id  = boundary_scope.org.id
  policy_id = boundary_storage_policy.boo.id
}
```

We can always migrate this once we have more information about future policies and how those would be assocaited with other resources.

internal/provider/provider.go
```diff
@@ -120,6 +120,8 @@ func New() *schema.Provider {
 			"boundary_host_set":                                 resourceHostSet(),
 			"boundary_host_set_static":                          resourceHostSetStatic(),
 			"boundary_host_set_plugin":                          resourceHostSetPlugin(),
+			"boundary_storage_policy":                           resourcePolicyStorage(),
```
louisruch, line 123: This should be `boundary_policy_storage`

internal/provider/resource_policy_storage.go
```diff
@@ -0,0 +1,376 @@
...
+			},
+			ScopeIdKey: {
+				Description: "The scope for this policy.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			policyStorageRetainForDaysKey: {
+				Description: "The number of days a session recording is required to be stored. Defaults to 0: allow deletions at any time. However, " + policyStorageRetainForDaysKey + " and " + policyStorageDeleteAfterDaysKey + " cannot both be 0.",
+				Type:        schema.TypeInt,
+				Optional:    true,
```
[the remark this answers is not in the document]
kheina: good catch

internal/provider/resource_policy_attachment.go
kheina: while waiting for adam's thoughts, my response was going to be that having the policy variable be `policy_id = boundary_policy_storage.foo.id` would make it clear enough what type of policy you're attaching. in the mean time I've changed the resource to `boundary_scope_policy_attachment`

AdamBouhmad: I think personally I prefer the former approach, given that storage policies are a subset of governance policies, and a scope in the future may have multiple policies associated((e.g a future policy could enforce session recording at the org level, and have that be enforced on all SSH targets in child scopes). There's a very similar pattern with [terraform iam policies](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_user_policy_attachment), where you'd define your iam policies, and attach them to a particular user

example:

`resource "aws_iam_user_policy_attachment" "s3" {
  user       = aws_iam_user.example_user.name
  policy_arn = aws_iam_policy.policy_s3.arn
}

resource "aws_iam_user_policy_attachment" "dynamodb" {
  user       = aws_iam_user.example_user.name
  policy_arn = aws_iam_policy.policy_dynamodb.arn
}`

louisruch (approved):
This looks great! :shipit:

PR #571, opened 2024-03-02
Add target alias resource

talanknight:


louisruch (commented):
This looks good, however tests are failing because we are spinning up a version of Boundary without alias support it seems.

louisruch:
Also I made a couple updates to deps so you might need to rebase

louisruch (approved):
:shipit:

PR #613, opened 2024-05-09
chore: Generate datasources based on OpenAPI specification

moduli:
NOTE: WIP

This PR is a continuation of the work started here: https://github.com/hashicorp/terraform-provider-boundary/pull/127

...

To Dos
- [ ] Review tests added from original PR
- [ ] Add tests for new data sources that were created
- [ ] Manual testing
- [ ] Generate examples for use in docs
- [ ] Figure out if we also want to include "singular" variants (i.e. `data "boundary_auth_method"` v.s. `data "boundary_auth_methods"`)


internal/provider/data_source_targets.go
moduli, line 1: The schema here is going to make documentation a little tricky since not all of this information is returned in a LIST response...

PR #617, opened 2024-07-10
chore: replace deprecated arguments in provider example

bschaatsbergen:
The `password_auth_method_login_name` and `password_auth_method_password` arguments are deprecated in favor of `auth_method_login_name` and `auth_method_password`.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,23 +15,23 @@ Do not keep your authentication password in HCL for production environments, use
 
 ```terraform
 provider "boundary" {
-  addr                   = "http://127.0.0.1:9200"
-  auth_method_id         = "ampw_1234567890" # changeme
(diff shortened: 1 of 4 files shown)


docs/index.md
moduli, line 1: Instead of modifying this file directly, I believe you need to modify `examples/provider/provider.tf` first and then run `make docs` to regenerate this file.

moduli (changes requested):
Thanks for catching this.

docs/index.md
bschaatsbergen: Thanks @moduli, fixed

moduli: @bschaatsbergen Looks good.

We may want to also update the example in `README.md`? Looks like there's an instance there.

moduli (approved):

PR #619, opened 2024-07-17
docs: separate worker and controller led resources

bschaatsbergen:
- **What**: Separated the worker and controlled led `boundary_worker` into two distinct examples.
- **Why**: Combining both the controller and worker LED resources into a single example in the documentation can be confusing.


templates/resources/worker.md.tmpl
```diff
@@ -0,0 +1,26 @@
...
+subcategory: ""
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# boundary_worker (Resource)
+
+{{ .Description | trimspace }}
+
+## Controller-led worker
```
moduli, line 12: Suggestion: What if we organized it to have
```
## Example Usage

### Controller-led worker

### Worker-led worker
```
?

examples/resources/boundary_worker/controller_led.tf
```diff
@@ -0,0 +1,5 @@
+resource "boundary_worker" "controller_led" {
+  scope_id    = "global"
+  name        = "worker 1"
```
moduli, line 3: Should we give the workers different names to reinforce the idea that resources need to have unique names?

bschaatsbergen: Good catch. I’ll change it to “controller-led-worker-1” and “worker-led-worker-1”.​

templates/resources/worker.md.tmpl
bschaatsbergen: Fixed 👌🏼

moduli (approved):

PR #626, opened 2024-10-01
feat(provider): Support worker filter on dynamic host catalogs

hugoghx:


hugoghx:
TODO: Need to update to Boundary next to solve test issues
```
panic: handlers.NewMaskManager: mapping src field "worker_filter" maps to "WorkerFilter", dest "WorkerFilter" maps to "": encoding issue: error #302
```

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.23.1
 require (
 	github.com/YakDriver/regexache v0.23.0
 	github.com/hashicorp/boundary v0.17.1
-	github.com/hashicorp/boundary/api v0.0.52
-	github.com/hashicorp/boundary/sdk v0.0.48
+	github.com/hashicorp/boundary/api v0.0.53
```
louisruch, line 8: If you want we can update to boundary@main which will pull the current SHA. Then after we release we can make a smaller PR to just update that to 0.18.0, update the changelog date and release the TF provider

hugoghx: Sounds good! Updated with https://github.com/hashicorp/terraform-provider-boundary/pull/626/commits/acd3e53e4649bd53773321e87f96a4feb5f9a933

docs/resources/host_catalog_plugin.md
```diff
@@ -94,6 +94,7 @@ resource "boundary_host_catalog_plugin" "azure_example" {
 - `plugin_name` (String) The name of the plugin that should back the resource. This or plugin_id must be defined.
 - `secrets_hmac` (String) The HMAC'd secrets value returned from the server.
 - `secrets_json` (String, Sensitive) The secrets for the host catalog. Either values encoded with the "jsonencode" function, pre-escaped JSON string, or a file:// or env:// path. Set to a string "null" to clear any existing values. NOTE: Unlike "attributes_json", removing this block will NOT clear secrets from the host catalog; this allows injecting secrets for one call, then removing them for storage.
+- `worker_filter` (String) HCP Only. A filter used to control which PKI workers can handle dynamic host catalog requests.
```
louisruch, line 97: Interesting I wonder if we need to update our other usage of HCP Only - this is a relic of before we had self hosted. Perhaps we should align this with Enterprise-only as we did in Boundary. But perhpas worth starting a thread and taggin Anando from product on his thoughts

louisruch (commented):
LGTM for the most part one question on HCP Only vs Enterprise Only wording.

We will also need to include a changelog entry for the addition of worker filter

docs/resources/host_catalog_plugin.md
hugoghx: Yeah, I really only followed our current convention in these docs, but this is technically incorrect.

I suppose we could update it everywhere to say `HCP and Boundary Enterprise only` or something like this. @anando-chatterjee WDYT?

hugoghx: (Also added a CHANGELOG entry with https://github.com/hashicorp/terraform-provider-boundary/compare/9d5d0ec74ba5c995e2674612074382ac974d7f3f..fc7f471ebf2aaddfc965552813b0ab6bdd713f9d)

hugoghx: Updated to Boundary v0.18.0
https://github.com/hashicorp/terraform-provider-boundary/compare/fc7f471ebf2aaddfc965552813b0ab6bdd713f9d..4efc3734c0003906ab001cfdadb4e8924355ff51

louisruch (approved):
LGTM dont push a tag just yet

PR #627, opened 2024-10-08
chore(provider): Remove support for GrantScopeId

hugoghx:
With Boundary 0.15, a deprecation notice was put under the `grant_scope_id` field, and a new `grant_scope_ids` field was introduced to replace it.

With Boundary v0.17.1 and [Boundary API v0.0.52](https://github.com/hashicorp/boundary/commit/c6f3c375f39737fdec8324f155c1cba91daa6c55), `grant_scope_id` support was entirely removed. This PR removes `grant_scope_id` support from this TF provider.


louisruch (approved):
Thanks!

PR #700, opened 2025-05-21
feat(provider): Support UPD credentials

hugoghx:


hugoghx:
Ran acceptance tests as well as manual tests locally -- All passing.

These won't be able to run in GitHub due to the Boundary dependencies.

PR #753, opened 2025-10-14
feat: Add `is_primary_for_scope` attribute to `boundary_auth_method` and `boundary_auth_method_password` resources

### LINKED ISSUE #754: Add `is_primary_for_scope` attribute to `boundary_auth_method_password` resource
Currently, the `boundary_auth_method_password` resource does not have an attribute to mark the auth method as the primary method for its scope.

As a result, the only way to make a password auth method primary is to activate it outside of Terraform. Until then, when using Boundary via the CLI, the user must explicitly provide the auth method ID to authenticate, which is inconvenient and reduces the usability of the CLI.

**Potential Terraform Configuration**
```tf
resource "boundary_auth_method_password" "password" {
  scope_id = boundary_scope.org.id
  is_primary_for_scope = true
}
```

rayennh:
This PR adds a new attribute, `is_primary_for_scope`, for both the `boundary_auth_method` and `boundary_auth_method_password` resources.

```bash
resource "boundary_scope" "org" {
  name                     = "organization_one"
  description              = "My first scope!"
  scope_id                 = "global"
  auto_create_admin_role   = true
  auto_create_default_role = true
}

resource "boundary_auth_method_password" "password" {
  scope_id = boundary_scope.org.id
  is_primary_for_scope = true
}
```

**Notes**
I don't know what direction will be given to `boundary_auth_method` compared to `boundary_auth_method_password`. To be safe, I’ve added the new attribute to both resources for consistency.

<details>
<summary>Acceptance tests</summary>

```bash
$ TF_ACC=1 go test -run TestAccBaseAuthMethod -v ./internal/provider
=== RUN   TestAccBaseAuthMethodPassword
--- PASS: TestAccBaseAuthMethodPassword (8.68s)
=== RUN   TestAccBaseAuthMethodPasswordIsPrimary
--- PASS: TestAccBaseAuthMethodPasswordIsPrimary (13.13s)
PASS
ok      github.com/hashicorp/terraform-provider-boundary/internal/provider      22.773s

$ TF_ACC=1 go test -run TestAccAuthMethodPassword -v ./internal/provider 
=== RUN   TestAccAuthMethodPassword
--- PASS: TestAccAuthMethodPassword (9.52s)
=== RUN   TestAccAuthMethodPasswordIsPrimary
--- PASS: TestAccAuthMethodPasswordIsPrimary (12.90s)
PASS
ok      github.com/hashicorp/terraform-provider-boundary/internal/provider      23.504s
```
</details>

Closes #754

### DIFF
--- a/internal/provider/resource_auth_method_password.go
+++ b/internal/provider/resource_auth_method_password.go
@@ -265,10 +265,6 @@ func resourceAuthMethodPasswordUpdate(ctx context.Context, d *schema.ResourceDat
 			if err := updateScopeWithPrimaryAuthMethodId(ctx, scopeId, authMethodId, meta); err != nil {
 				return diag.Errorf("%v", err)
 			}
-		} else {
-			if err := updateScopeWithPrimaryAuthMethodId(ctx, scopeId, "", meta); err != nil {
-				return diag.Errorf("%v", err)
-			}
 		}
 	}
 
--- a/internal/provider/resource_auth_method.go
+++ b/internal/provider/resource_auth_method.go
@@ -273,10 +273,6 @@ func resourceAuthMethodUpdate(ctx context.Context, d *schema.ResourceData, meta 
 			if err := updateScopeWithPrimaryAuthMethodId(ctx, scopeId, authMethodId, meta); err != nil {
 				return diag.Errorf("%v", err)
 			}
-		} else {
-			if err := updateScopeWithPrimaryAuthMethodId(ctx, scopeId, "", meta); err != nil {
-				return diag.Errorf("%v", err)
-			}
 		}
 	}
 
--- a/internal/provider/resource_auth_method_oidc.go
+++ b/internal/provider/resource_auth_method_oidc.go
@@ -405,12 +405,7 @@ func updateScopeWithPrimaryAuthMethodId(ctx context.Context, scopeId, authmethod
 
 	opts := []scopes.Option{}
 	opts = append(opts, scopes.WithAutomaticVersioning(true))
-
-	if authmethodId != "" {
-		opts = append(opts, scopes.WithPrimaryAuthMethodId(authmethodId))
-	} else {
-		opts = append(opts, scopes.DefaultPrimaryAuthMethodId())
-	}
+	opts = append(opts, scopes.WithPrimaryAuthMethodId(authmethodId))
 
 	_, err := scp.Update(ctx, scopeId, 0, opts...)
 	if err != nil {
--- a/internal/provider/resource_auth_method_password_test.go
+++ b/internal/provider/resource_auth_method_password_test.go
@@ -45,16 +45,6 @@ resource "boundary_auth_method_password" "foo" {
 	is_primary_for_scope = true
 	depends_on  = [boundary_role.org1_admin]
 }`, fooAuthMethodDesc)
-
-	fooAuthMethodIsPrimaryUpdateWithOtherChange = fmt.Sprintf(`
-resource "boundary_auth_method_password" "foo" {
-	name        = "test"
-	description = "%s"
-	type        = "password"
-	scope_id    = boundary_scope.org1.id
-	is_primary_for_scope = true
-	depends_on  = [boundary_role.org1_admin]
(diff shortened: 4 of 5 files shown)


internal/provider/resource_auth_method_password.go
```diff
@@ -15,9 +15,10 @@ import (
 )
 
 const (
-	authmethodTypePassword          = "password"
-	authmethodMinLoginNameLengthKey = "min_login_name_length"
-	authmethodMinPasswordLengthKey  = "min_password_length"
+	authmethodTypePassword                   = "password"
+	authmethodMinLoginNameLengthKey          = "min_login_name_length"
+	authmethodMinPasswordLengthKey           = "min_password_length"
+	authmethodIsPrimaryAuthMethodForScopeKey = "is_primary_for_scope"
```
moduli: Wondering if this could be shortened to just `authmethodIsPrimaryForScopeKey`

internal/provider/resource_auth_method_password.go
```diff
@@ -177,6 +200,13 @@ func resourceAuthMethodPasswordRead(ctx context.Context, d *schema.ResourceData,
 		return diag.Errorf("auth method nil after read")
 	}
 
+	serr, isPrimary := readScopeIsPrimaryAuthMethodId(ctx, amrr.GetResponse().Map["scope_id"].(string), amrr.GetResponse().Map["id"].(string), meta)
+	if serr != nil {
+		return diag.Errorf("%v", serr)
+	}
+
+	amrr.GetResponse().Map[authmethodIsPrimaryAuthMethodForScopeKey] = isPrimary
```
moduli: Noticing in the other auth method files, this was
```
if isPrimary {
		amrr.GetResponse().Map[authmethodOidcIsPrimaryAuthMethodForScope] = true
	}
```

What was the reason for the change? I get that it might be equivalent, but is it worth to keep things consistent?

internal/provider/resource_auth_method_password.go
```diff
@@ -218,15 +248,47 @@ func resourceAuthMethodPasswordUpdate(ctx context.Context, d *schema.ResourceDat
...
+		}
+
 		return setFromPasswordAuthMethodResponseMap(d, amur.GetResponse().Map)
 	}
+
+	// If only is_primary_for_scope changed
+	if d.HasChange(authmethodIsPrimaryAuthMethodForScopeKey) {
+		return resourceAuthMethodPasswordRead(ctx, d, meta)
+	}
+
```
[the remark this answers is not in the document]
moduli: Looks like one difference in the implementation is to handle a case where `is_primary_for_scope` is the only change in the update? Is there a reason why that was only handled in `if len(opts) > 0 {` before?

internal/provider/resource_auth_method_password.go
rayennh: If there was a specific reason I can not remember it so I updated the code to match the OIDC/LDAP implementation

internal/provider/resource_auth_method_password.go
```diff
@@ -218,15 +248,47 @@ func resourceAuthMethodPasswordUpdate(ctx context.Context, d *schema.ResourceDat
...
+			return diag.Errorf("error updating auth method: nil resource")
+		}
+		scopeId := amrr.GetResponse().Map["scope_id"].(string)
+		authMethodId := amrr.GetResponse().Map["id"].(string)
+
+		if d.Get(authmethodIsPrimaryAuthMethodForScopeKey).(bool) {
+			if err := updateScopeWithPrimaryAuthMethodId(ctx, scopeId, authMethodId, meta); err != nil {
+				return diag.Errorf("%v", err)
+			}
+		} else {
```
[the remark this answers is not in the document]
louisruch: I would want to be consistent in how we approach this for all scenarios. If we want to allow unsettling - I would need to chat with the rest of the team how we want to approach this and we would need to update oidc/ldap.

Overall I am happy with this PR, my recommendation is that we remove the else block and then merge it. We can then, if we feel it is the correct approach, open a different PR to update the behavior of all oidc/ldap/password

PR #838, opened 2026-09-08
Supress group name case sensitive matching

abhi99shek99:
## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

### DIFF
--- a/internal/provider/resource_managed_group_ldap.go
+++ b/internal/provider/resource_managed_group_ldap.go
@@ -6,6 +6,7 @@ package provider
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/managedgroups"
@@ -56,6 +57,9 @@ func resourceManagedGroupLdap() *schema.Resource {
 					Type: schema.TypeString,
 				},
 				Required: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return strings.EqualFold(old, new)
+				},
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


SunithaGudisagarIBM1 (dismissed):

internal/provider/resource_managed_group_ldap.go
```diff
@@ -56,6 +57,9 @@ func resourceManagedGroupLdap() *schema.Resource {
 					Type: schema.TypeString,
 				},
 				Required: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return strings.EqualFold(old, new)
+				},
```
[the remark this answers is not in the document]
abhi99shek99: Done

internal/provider/resource_managed_group_ldap.go
```diff
@@ -52,10 +53,13 @@ func resourceManagedGroupLdap() *schema.Resource {
 			managedGroupLdapGroupNamesKey: {
 				Description: "The list of groups that make up the managed group.",
 				Type:        schema.TypeList,
+				Required:    true,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
+					DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
anilvpatel, line 59: In Terraform Plugin SDK v2, DiffSuppressFunc is officially supported only on top-level schema.Schema fields, not on Elem schemas nested within a TypeList. If the SDK silently ignores this, the entire purpose of this change (suppressing case-only diffs) would be lost. The correct approach is to move the DiffSuppressFunc to the parent TypeList schema field itself, or normalize casing when writing values back to state in setFromManagedGroupLdapResponseMap.

abhi99shek99: Done

anilvpatel (approved):

Issue #664, opened 2025-01-22
[Feature Request] Add a resource "boundary_target_host_source"


lde-axelor:
Please add a "boundary_target_host_source" to be able to define host sources for an existing target.

### Use case

Having two different terraform repository : One to define all infra resources of a project and one to define all project resources.

Infra resources includes for example a boundary target for the database without the address because the database address doesn't exist yet at this stage.

Project resources includes the DNS of the database in a dedicated boundary host source, which is known at this stage because database is a project resource and which change on each redeployments.

This would allow us to define fine tuned permission grants on targets on infra side while being able to define target addresses dynamically on project sides.

PR #839, opened 2026-09-16
feat(provider): add boundary_target_host_source resource

andrewkim099:
## Description
This addresses Issue #664, also listed in https://hashicorp.atlassian.net/browse/ICU-20176.

Add `boundary_target_host_source` to manage target-to-host-source attachments independently from `boundary_target.host_source_ids`.

## Changes

- register the new resource in the provider
- implement create/read/delete logic with idempotent checks
- add acceptance tests for create, import, and removal

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
Not applicable

- [x] If applicable, I've documented the impact of any changes to security controls.
Not applicable

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.


internal/provider/resource_target_host_source.go
```diff
@@ -0,0 +1,191 @@
...
+
+func resourceTargetHostSourceCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	md := meta.(*metaData)
+	tc := targets.NewClient(md.client)
+
+	targetId := d.Get(targetHostSourceTargetIdKey).(string)
+	hostSourceId := d.Get(targetHostSourceIdKey).(string)
+
+	// Read the target to get its current host source list.
+	trr, err := tc.Read(ctx, targetId)
```
matt-romanelli, line 64: I'd recommend against abbreviations as variables like md, tc, and trr as they are less intuitive. err is an exception as that is Go convention to use for errors. But if any of the other abbreviations are accepted practice, please correct me!

Is trr supposed to be the target object? If so, use 'target'.

internal/provider/resource_target_host_source.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+const (
+	targetHostSourceTargetIdKey = "target_id"
+	targetHostSourceIdKey       = "host_source_id"
+)
+
+func resourceTargetHostSource() *schema.Resource {
+	return &schema.Resource{
+		Description: "Attaches a host source to an existing Boundary target. " +
```
matt-romanelli: Could this, along with the other descriptions, be made a constant?

internal/provider/resource_target_host_source.go
```diff
@@ -0,0 +1,191 @@
...
+
+	if _, err := tc.RemoveHostSources(ctx, targetId, 0, []string{hostSourceId}, targets.WithAutomaticVersioning(true)); err != nil {
+		return diag.Errorf("error removing host source from target: %v", err)
+	}
+
+	return nil
+}
+
+// targetAndHostSourceIdsFromResourceData returns the target ID and host source ID
+// from either the resource attributes (normal CRUD) or the composite resource ID (import).
```
matt-romanelli, line 174: CRUD is mentioned but I only see operations for Create, Read and Delete. Is Update not needed?

matt-romanelli (commented):
Overall the code is very clean and well written. 

But how was this feature tested to confirm that it works as expected? The unit tests provide good coverage, especially to prevent against future breaking changes, but I'd like to hear how you manually tested this to confirm that the feature has been fully implemented.

internal/provider/resource_target_host_source.go
andrewkim099: Thanks for the comment. I was trying to match the syntax in resource_target.go which uses these abbreviations. Should I still change it?

internal/provider/resource_target_host_source.go
andrewkim099: Update is not needed because both fields in the schema are ForceNew so they can't be updated

internal/provider/resource_target_host_source.go
matt-romanelli: If this syntax already exists and is extensively used then I agree to keep it consistent. If it doesn't have much usage elsewhere, then this could be a good opportunity to make the code more readable. 

I'm pretty new to Go so maybe this naming structure is common.

matt-romanelli (approved):

